PR #9, opened 2021-10-15
Improving tests

Abuelodelanada:
This PR fixes:
- Loki tests follows the same pattern used for Prometheus
- `test_remote_write_configuration` was broken. Sometimes returned green and sometimes red because of lists inside dictionaries
- Add `-s` parameter in pytest execution 

Related to OP-370

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -234,16 +234,25 @@ class TestCharm(unittest.TestCase):
         expected = "-config.file=/etc/agent/agent.yaml -prometheus.wal-directory=/tmp/agent/data"
         self.assertEqual(self.harness.charm._cli_args(), expected)
 
-    def test__loki_config_empty(self):
-        self.harness.charm._stored.remove_loki_config = True
-        self.assertEqual(self.harness.charm._loki_config(), {})
-
-        self.harness.charm._stored.remove_loki_config = False
-        self.assertEqual(self.harness.charm._loki_config(), {})
-
-    def test__loki_config_non_empty(self):
+    @responses.activate
+    @patch.object(Container, "pull", new=pull_empty_fake_file)
+    @patch.object(Container, "push")
+    def test__on_logging_relation_changed(self, mock_push: MagicMock):
         self.harness.charm._loki_consumer = Mock()
         self.harness.charm._loki_consumer.loki_push_api = "http://loki:3100:/loki/api/v1/push"
+        mock_push.push.return_value = None
+
+        responses.add(
+            responses.POST,
+            "http://localhost/-/reload",
+            status=200,
+        )
+
+        rel_id = self.harness.add_relation("logging", "consumer")
+        self.harness.add_relation_unit(rel_id, "consumer/0")
+        self.harness.update_relation_data(rel_id, "consumer/0", {})
+
+        path, content = mock_push.call_args[0]
 
         expected = {
             "configs": [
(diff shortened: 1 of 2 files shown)


tests/unit/test_charm.py
```diff
@@ -116,35 +117,65 @@ def test_remote_write_configuration(self, mock_push: MagicMock):
...
+        #     {"url": "http://1.1.1.1:9090/api/v1/write"},
+        # ],
+        #
+        # and there is no guarantee that the lists will come in the same order every time
+        # we run the tests, we cannot use `self.assertDictEqual()` and we have
+        # to compare the dictionaries in parts.
+        self.assertDictEqual(
+            content["integrations"]["agent"], expected_config["integrations"]["agent"]
+        )
+        self.assertTrue(
```
sed-i, line 162: Consider using `set()` methods or `filter`.

tests/unit/test_charm.py
```diff
@@ -116,35 +117,65 @@ def test_remote_write_configuration(self, mock_push: MagicMock):
...
-                        {
-                            "name": "agent_scraper",
-                            "remote_write": [
-                                {"url": "http://1.1.1.2:9090/api/v1/write"},
-                                {"url": "http://1.1.1.1:9090/api/v1/write"},
-                            ],
-                            "scrape_configs": [],
-                        }
-                    ]
+        content = yaml.load(content, Loader=yaml.FullLoader)
```
sed-i, line 120: why not `yaml.safe_load`?

tests/unit/test_charm.py
```diff
@@ -116,35 +117,65 @@ def test_remote_write_configuration(self, mock_push: MagicMock):
...
+        # for instance:
+        #
+        # "remote_write": [
+        #     {"url": "http://1.1.1.2:9090/api/v1/write"},
+        #     {"url": "http://1.1.1.1:9090/api/v1/write"},
+        # ],
+        #
+        # and there is no guarantee that the lists will come in the same order every time
+        # we run the tests, we cannot use `self.assertDictEqual()` and we have
+        # to compare the dictionaries in parts.
```
sed-i, line 158: Maybe the value of this test would be increased if broken down into smaller pieces? Not sure.

tests/unit/test_charm.py
```diff
@@ -238,63 +278,33 @@ def test__loki_config_non_empty(self):
...
-        self.assertDictEqual(yaml.safe_load(self.harness.charm._config_file()), expected)
+        self.assertDictEqual(yaml.safe_load(content)["loki"], expected)
+        self.assertEqual(path, "/etc/agent/agent.yaml")
+        self.assertEqual(self.harness.model.unit.status, ActiveStatus())
 
-    def test__config_file_with_loki(self):
+    @responses.activate
+    @patch.object(Container, "pull", new=pull_empty_fake_file)
+    @patch.object(Container, "push")
+    def test__on_logging_relation_departed(self, mock_push: MagicMock):
```
sed-i, line 288: Could you name/document the test to be more immediately obvious what it is testing? Applies to some of the other tests as well.

tests/unit/test_charm.py
Abuelodelanada: I didn't modify the way the dictionary was loaded, I only assign it to a variable.
I'll change it.

tests/unit/test_charm.py
sed-i: even better

tests/unit/test_charm.py
Abuelodelanada: Sure, I always prefer little tests anyway

PR #12, opened 2021-10-19
Update Loki library

Abuelodelanada:
In this PR we:

- Update the Loki library
- Refactor charm code to use `LokiPushApiEndpointDeparted` and `LokiPushApiEndpointJoined`
- Add tests

### DIFF
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -231,7 +231,6 @@ class TestCharm(unittest.TestCase):
                                     "http_listen_port": 3500,
                                     "grpc_listen_port": 3600,
                                 },
-                                "labels": {"pushserver": "loki"},
                             },
                         }
                     ],


src/charm.py
```diff
@@ -180,21 +188,23 @@ def _cli_args(self) -> str:
...
         if integrations_config := self._integrations_config():
             config["integrations"] = integrations_config
+
         if prometheus_config := self._prometheus_config():
             config["prometheus"] = prometheus_config
-        if loki_config := self._loki_config():
-            config["loki"] = loki_config
+
+        self._loki_config(config, event)
         return yaml.dump(config)
```
sed-i, line 208: General comment: from first read it is difficult for me to follow what is going on (how a config is constructed), but it may be due to the complexity involved.

src/charm.py
```diff
@@ -274,24 +284,21 @@ def _prometheus_config(self) -> dict:
...
+            config.pop("loki", None)
 
-        if self._stored.remove_loki_config:
-            self._stored.remove_loki_config = False
-            return {}
-
-        if loki_push_api := self._loki_consumer.loki_push_api:
-            config = {
+        if isinstance(event, LokiPushApiEndpointJoined):
+            config["loki"] = {
```
sed-i, line 297: Consider recreating the entire config based on the remaining relation, instead of using isinstance and differential config update.
Also it looks like "_loki_config" could be moved into the loki lib?

src/charm.py
```diff
@@ -151,13 +157,15 @@ def _patch_k8s_service(self):
                 self._stored.k8s_service_patched = True
                 logger.info("Successfully patched the Kubernetes service!")
 
-    def _update_config(self):
+    def _update_config(self, event=None):
```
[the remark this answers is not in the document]
Abuelodelanada: There are two calls to `_update_config()` without an event arg: 
- [L116](https://github.com/canonical/grafana-agent-operator/blob/7502d3818a8404a67fe5a2c365beb3d98625af7f/src/charm.py#L116)
- [L121](https://github.com/canonical/grafana-agent-operator/blob/7502d3818a8404a67fe5a2c365beb3d98625af7f/src/charm.py#L121)

src/charm.py
```diff
@@ -180,21 +188,23 @@ def _cli_args(self) -> str:
...
+    def _config_file(self, event: EventBase) -> str:
+        """Generates config file str based on the event received.
 
         Returns:
-            The config file contents
+            A yaml string with grafana agent config
         """
         config = {}
         if server_config := self._server_config():
             config["server"] = server_config
```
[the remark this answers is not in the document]
Abuelodelanada: Yes, it's nicer!

src/charm.py
Abuelodelanada: Yes, you are right, it's not clear. I'll refactor it.

PR #16, opened 2021-11-09
First approach. Downloading promtail and uploading to workload container

Abuelodelanada:
- [x] Provider side of the relationship sends loki_push_api URL over relation data
- [x] Provider side of the relationship send URL to download Promtail binary over relation data
- [x] Consumer side builds Promtail config and push it to the container
- [x] Consumer side downloads the binary if the binary is not in the filesystem
- [x] Consumer side build Pebble layer for Promtail execution and starts this service
- [ ] Promtail binary should be uploaded to charmhub as a Loki resource


lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -97,9 +102,19 @@ def __init__(self, *args):
...
-CONFIG_PATH = "/etc/promtail/config.yml"
+CONFIG_PATH = "/tmp/promtail_config.yml"
+PROMTAIL_BINARY_ZIP_URL = "https://github.com/grafana/loki/releases/download/v2.4.1/promtail-linux-amd64.zip"
+BINARY_ZIP_PATH = "/tmp/promtail-linux-amd64.zip"
+BINARY_DIR = "/tmp"
+BINARY_SHA25SUM = "978391a174e71cfef444ab9dc012f95d5d7eae0d682eaf1da2ea18f793452031"
+WORKLOAD_BINARY_PATH = "/tmp/promtail-linux-amd64"
+
+
+class PromtailSHA256Error(Exception):
```
mmanciop, line 116: Just call it `PromtailDigestError`, there is no reason to hard-code the hashing scheme in the exception name

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -127,9 +141,10 @@ def __init__(self, charm, log_files: list, relation_name: str = DEFAULT_RELATION
         self._stored.set_default(grafana_agents="{}")
         self._charm = charm
         self._relation_name = relation_name
+        containers = dict(self._charm.model.unit.containers)
+        self._container_name = [*containers].pop()
```
mmanciop, line 145: This dangerously assumes that the container we care about is the first. This is not a healthy assumption unless there is exactly ONE container. Please add the container name as an optional argument, and is `None` is provided, check if there is just ONE container and if not fail hard.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -177,15 +192,53 @@ def _on_promtail_pebble_ready(self, _):
                 "promtail": {
                     "override": "replace",
                     "summary": "promtail",
-                    "command": f"/usr/bin/promtail {self._cli_args}",
-                    "startup": "disable",
+                    "command": f"{WORKLOAD_BINARY_PATH} {self._cli_args}",
```
mmanciop, line 195: This is not Python 3.5 compliant

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -79,10 +79,15 @@ def __init__(self, *args):
 
 import logging
 import json
+import requests
```
mmanciop, line 82: Can't we use `http.client`? `requests` becomes a dependency people must add in their `requirements.txt`

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -97,9 +102,19 @@ def __init__(self, *args):
...
 HTTP_LISTEN_PORT = 9080
+HTTP_LISTEN_PORT = 9080
 GRPC_LISTEN_PORT = 0
 POSITIONS_FILENAME = "/tmp/positions.yaml"
-CONFIG_PATH = "/etc/promtail/config.yml"
+CONFIG_PATH = "/tmp/promtail_config.yml"
+PROMTAIL_BINARY_ZIP_URL = "https://github.com/grafana/loki/releases/download/v2.4.1/promtail-linux-amd64.zip"
+BINARY_ZIP_PATH = "/tmp/promtail-linux-amd64.zip"
+BINARY_DIR = "/tmp"
+BINARY_SHA25SUM = "978391a174e71cfef444ab9dc012f95d5d7eae0d682eaf1da2ea18f793452031"
```
mmanciop, line 112: I'd locate this and `PROMTAIL_BINARY_ZIP_URL` close to the `LIBPATCH` part, so that we have one place with all the weird bits to update between versions :D

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -254,18 +307,15 @@ def _build_config_file(self, event) -> str:
 
         return yaml.dump(config)
 
-    def _initial_config(self) -> dict:
+    def _initial_config(self) -> None:
```
mmanciop, line 310: The name of this method really does not suggest you are pushing stuff.

I'd keep `_initial_config` as a property, and do the pushing somewhere else.

mmanciop (changes requested):

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
jnsgruk: (I'm gonna review this more later)

I don't think this matters? This charm code will run on Python 3.8, no?

mmanciop: True that :-)

mmanciop: However for this particular charm, I see a very likely chance that we will want to make a machine subordinate version, hence Python 3.5 overall sounds to me like it makes sense.

mmanciop: @Abuelodelanada besides, I thought we said this library would actually be owned by the Loki charm.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: Sure, I harcoded this to the first container because I wanted something that just work to continue with the development.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: @mmanciop what I thought we said was the "binary" would be as a Loki resource..... Are you sure you want this in Loki? I thought that our "entrypoint" for logs (and metrics) in LMA will be `grafana-agent`

simskij: > I thought we said this library would actually be owned by the Loki charm.

Yes, we did that, and I personally still think that's a good idea.

Abuelodelanada: Ok, it's not a big deal move the code to Loki.
So, should we provide Loki URL instead of Grafana-agent one to the consumer, right?

mmanciop: If you are wondering about the fact that Loki will not use `promtail` ever as a resource, worry you not: how do you think you are going to implement the logging part of self-monitoring for loki-k8s? 😈

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -155,37 +192,88 @@ def _on_log_proxy_relation_departed(self, event):
...
         }
+        self._container.add_layer(self._container_name, pebble_layer, combine=True)
+
+    def _obtain_promtail(self, event) -> None:
+        if self._is_promtail_binary_in_workload():
+            return
+
+        self._download_promtail(event)
+
+        if not self._check_sha256sum():
```
jnsgruk, line 237: I'd maybe just combine `_download_promtail` and `self._check_sha256sum` and just have the `_download_promtail` method raise the `PromtailDigestError` if the checksum is off? It seems the `_check_sha256sum` is not called anywhere else so probably not that useful in its own right?

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -155,37 +192,88 @@ def _on_log_proxy_relation_departed(self, event):
...
+            msg = "Promtail bnary sha256sum mismatch"
+            logger.error(msg)
+            raise PromtailDigestError(msg)
+
+        self._unzip_binary()
```
jnsgruk, line 242: Again, I think the `_download_promtail` method should probably just take care of this, not really sure why it's a distinct step? Are we eveer going to do anything *other than* want to extract the binary?

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -155,37 +192,88 @@ def _on_log_proxy_relation_departed(self, event):
...
+        self._upload_binary()
```
jnsgruk, line 243: The number of these little 1-2 line helpers makes this method kinda awkward to read imo -- consider the following:

```python
from shutil import copyfileobj

def _download_promtail(self, event) -> None:
    # Fetch the binary
    response = urlopen(PROMTAIL_BINARY_ZIP_URL)
    with response as r, open(BINARY_ZIP_PATH, "wb") as f:
        copyfileobj(r, f)

    # Validate that it's legit
    with open(BINARY_ZIP_PATH, "rb") as f:
        result = sha256(f.read()).hexdigest()

    if BINARY_SHA256SUM != result:
        logger.error(
            "promtail binary mismatch, expected:'{}' but got '{}', 
            BINARY_SHA256SUM, result)
        raise PromtailDigestError("Digest mismatch for promtail binary")

    # Unzip the binary
    with ZipFile(BINARY_ZIP_PATH, "r") as z:
        z.extractall(BINARY_DIR)

    return "/tmp/promtail-linux-amd64"
    

def _obtain_promtail(self, event) -> None:
    if self._is_promtail_binary_in_workload():
        return
    
    bin_path = self._download_promtail(event)
    
    with open(bin_path, "rb") as f:
        self._container.push(WORKLOAD_BINARY_PATH, f, permissions=0o755)
```

You could even be a little racier, spend a bit of memory and reduce the amount of file IO you're doing, though this does assume at least ~75MB of free memory -- note however that right now Pebble uploads straight from memory either way, so if this approach isn't suitable, we may have wider problems:

```python
def _download_promtail(self, event) -> None:
  with urlopen(PROMTAIL_BINARY_ZIP_URL) as r:
      file_bytes = r.read()
      
      if sha256(file_bytes).hexdigest() != BINARY_SHA256SUM:
          logger.error(
            "promtail binary mismatch, expected:'{}' but got '{}'", 
            BINARY_SHA256SUM, result
          )
          raise PromtailDigestError("Digest mismatch for promtail binary")
      
      ZipFile(io.BytesIO(file_bytes)).extractall(BINARY_DIR)

  return "/tmp/promtail-linux-amd64"
```

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -155,37 +192,88 @@ def _on_log_proxy_relation_departed(self, event):
...
 
-        self._container.add_layer("promtail", pebble_layer, combine=True)
-        config = self._initial_config()
-        self._container.push(CONFIG_PATH, yaml.dump(config))
+    def _is_promtail_binary_in_workload(self) -> bool:
+        pat = WORKLOAD_BINARY_PATH.split("/")[-1]
+        return True if len(self._container.list_files(BINARY_DIR, pattern=pat)) == 1 else False
+
+    def _download_promtail(self, event) -> None:
+        url = json.loads(event.relation.data[event.unit].get("data"))["promtail_binary_zip_url"]
```
jnsgruk, line 250: Why is this URL not using the file constant?

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -155,37 +192,88 @@ def _on_log_proxy_relation_departed(self, event):
...
+        response = urlopen(url)
+
+        with open(BINARY_ZIP_PATH, "wb") as f:
+            while True:
+                chunk = response.read(8192)
+                if not chunk:
+                    break
+                f.write(chunk)
```
jnsgruk, line 258:
```suggestion
        response = urlopen(PROMTAIL_BINARY_ZIP_URL)
        with response as r, open(BINARY_ZIP_PATH, "wb") as f:
            copyfileobj(r, f)
```

jnsgruk: If you don't use the above impl, this can still be tidied.

jnsgruk (changes requested):

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -94,12 +99,24 @@ def __init__(self, *args):
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 LIBPATCH = 1
+PROMTAIL_BINARY_ZIP_URL = (
+    "https://github.com/grafana/loki/releases/download/v2.4.1/promtail-linux-amd64.zip"
+)
+BINARY_ZIP_PATH = "/tmp/promtail-linux-amd64.zip"
+BINARY_DIR = "/tmp"
+BINARY_SHA25SUM = "978391a174e71cfef444ab9dc012f95d5d7eae0d682eaf1da2ea18f793452031"
+WORKLOAD_BINARY_PATH = "/tmp/promtail-linux-amd64"
```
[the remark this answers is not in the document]
mmanciop: I have seen a lot of silly with tmp folders, some even in containers: mounting them over nfs (!), cron jobs to wipe them (!!!), security policies that disable executing processes from it (???).

we want the library to work also in machine charms written with OF. I think it is wholly legitimate to create a proper folder for it.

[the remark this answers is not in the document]
mmanciop: I like the /opt/promtail path btw

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -115,35 +132,55 @@ class RelationManagerBase(Object):
...
 
         Args:
             event: The event object `RelationChangedEvent`.
         """
         if event.relation.data[event.unit].get("data", None):
+            self._obtain_promtail(event)
             self._update_config(event)
             self._update_agents_list(event)
+            self._build_pebble_layer()
             self._container.restart(self._container_name)
```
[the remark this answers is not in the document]
Abuelodelanada: I was restarting the consumer side but it's a bug. I changed (not committed yet) this line to:

```python
self._container.restart(SERVICE_NAME)
```
Where: 

```python
SERVICE_NAME = "promtail"
```

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -115,35 +132,55 @@ class RelationManagerBase(Object):
...
+        """Event handler for the `log_proxy_relation_created`."""
+        self._container.push(CONFIG_PATH, yaml.dump(self._initial_config))
+
     def _on_log_proxy_relation_changed(self, event):
         """Event handler for the `log_proxy_relation_changed`.
 
         Args:
             event: The event object `RelationChangedEvent`.
         """
         if event.relation.data[event.unit].get("data", None):
```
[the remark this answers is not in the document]
Abuelodelanada: We run this (at this point) just to have a initial config created to populate it later with the Grafana Agent URLs when a relation is fully established.

We can refactor it, checking if `data` is `None` and then create the config... let me think if there is a better approach here.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
```diff
@@ -155,37 +192,88 @@ def _on_log_proxy_relation_departed(self, event):
...
-            self._container.restart(self._container_name)
+            self._container.restart("promtail")
 
     def _on_upgrade_charm(self, event):
         # TODO: Implement it ;-)
         pass
 
-    def _on_promtail_pebble_ready(self, _):
-        """Event handler for the pebble ready event.
+    def _get_container(self, container_name):
```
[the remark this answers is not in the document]
Abuelodelanada: I changed this implementation (not yet committed).

For instance in the consumer side you can send the container_name

```python
self._log_proxy = LogProxyConsumer(charm=self, log_files=LOG_FILES, container_name="workload_app")
```
And in the lib:

```python
    def __init__(
        self,
        charm,
        log_files: list,
        container_name: Optional[str],
        relation_name: str = DEFAULT_RELATION_NAME,
    ):
    ...

    def _get_container(self, container_name):
        if container_name is not None:
            return self._charm.unit.get_container(container_name)

        containers = dict(self._charm.model.unit.containers)

        if len(containers) == 1:
            return self._charm.unit.get_container([*containers].pop())

        # FIXME: Use custom exception
        raise Exception("Container cannot be obtained")
```

The idea is: 
- If you send the container_name, we'll use that name to get the container.
- If you do not send the container_name:
  - If there is only one container in the Pod, we'll use it
  - If there is more than one container we'll raise an Exception
  
  What do you think about this?

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: Yes, we are not using `_chech_sha256sum` outside this scope. Makes sense. Thanks.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: I think the method that it's doing these tasks is `_obtain_promtail()`
Right now, the method:

- Check if binary is already in the workload container, if not:
  - Download the binary
  - Check the SHA256 sum
  - Unzip the binary
  - Upload the binary to the workload container

```python
    def _obtain_promtail(self, event) -> None:
        if self._is_promtail_binary_in_workload():
            return

        self._download_promtail(event)

        if not self._check_sha256sum():
            msg = "Promtail bnary sha256sum mismatch"
            raise PromtailDigestError(msg)

        self._unzip_binary()
        self._upload_binary()
```

Do you think we should have a different approach?

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: My first approach was like this one, having everything in a single method: `_download_promtail()`, but I changed to the actual approach because the method had more than one responsibility (download the file, check the sha256sum, and unzip it)
Anyway I have no strong opinions about both approaches.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: We are using the constant in the provider side where we are setting the URL in relation data:

```python
    def _on_log_proxy_relation_changed(self, event):
        if event.relation.data[self._charm.unit].get("data") is None:
            data = {}
            data.update(json.loads(self._loki_push_api))
            data.update(json.loads(self._promtail_binary_url))
            event.relation.data[self._charm.unit].update({"data": json.dumps(data)})
...

    @property
    def _promtail_binary_url(self) -> str:
        # FIXME: Use charmhub's URL
        return json.dumps({"promtail_binary_zip_url": PROMTAIL_BINARY_ZIP_URL})
```

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
Abuelodelanada: Yes, I wanted to write the file to disk while downloading it to reduce the amount of RAM used... but your suggestion it's better because is simpler to read. Thanks!

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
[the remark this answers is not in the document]
Abuelodelanada: I missed the `# FIXME` comment there. I used `/tmp` just to move forward with the development and give me some room to think a better alternative to this.

Yes `/opt/promtail` it's a good option.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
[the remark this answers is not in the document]
mmanciop: That is what I was looking for, except the Exception message that provides no guidance whatsoever to understand what is wrong with it :-)

Go with:

```
No 'container_name' parameter has been specified; since this charmed operator is not running exactly one container, it must be specified which container to get logs from.
```

Note: there will be soon charms WITHOUT workload containers so the error message must account for the "0 containers" case.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
jnsgruk: In the current approach there is just an excessive amount of IO from all the opening/closing/reading/writing files. Interested in @simskij opinion here.

simskij: I get your point about the separation of concerns, @Abuelodelanada, but in this case I would suggest inlining your helper functions to make it a smoother read. If you also adopt @jnsgruk's suggestion with not writing the tarball to disk but rather store it in a file object,untar it in-memory and write the files you need to disk, the code will end up quite compact regardless.

lib/charms/grafana_agent_k8s/v0/grafana_agent.py
simskij: I think you can inline most of this to keep it a smoother read. While extracting is good and encouraged in general, it might make debugging and reading choppy by forcing you to jump in and out of the function for every line.

PR #17, opened 2021-11-09
Remote write 2

mmanciop:
Clean up the implementation of Prometheus Remote Write in Grafana Agent.


Abuelodelanada (dismissed):

lib/charms/prometheus_k8s/v0/prometheus_remote_write.py
```diff
@@ -634,9 +636,11 @@ def _set_endpoint_on_relation(self, relation: Relation) -> None:
 
         path = self._endpoint_path or ""
         if path and not path.startswith("/"):
-            path = f"/{path}"
+            path = str(path)
```
sed-i, line 639: iiuc the preceding `/` is still needed

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -313,7 +313,7 @@ def _on_scrape_targets_changed(self, event):
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 11
```
sed-i, line 316: prom scrape is now at `LIBPATCH = 12`. fetch-lib?

src/charm.py
```diff
@@ -124,6 +124,11 @@ def on_remote_write_changed(self, _: RelationChangedEvent) -> None:
         self._update_config()
         self._update_status()
 
+    def on_remote_write_broken(self, _: RelationBrokenEvent) -> None:
```
sed-i, line 127: This is unused.
If needed, then possibly we'd also need to register relation departed.

lib/charms/prometheus_k8s/v0/prometheus_scrape.py
sed-i: Still outdated.

sed-i (dismissed):
Note that both prom libs have PRs in review.

mmanciop:
For debugging purposes, we should annotate the Grafana Agent config with YAML comments how the IPs of the prometheus instances we remote-write to maps to their Juju identity:

```
integrations:
  prometheus_remote_write:
  - url: http://10.1.250.232:9090/api/v1/write  # Juju model X;Juju application Y; Juju unit Z
  - url: http://10.1.250.233:9090/api/v1/write # Juju model X;Juju application Y; Juju unit W
prometheus:
  configs:
  - name: agent_scraper
    remote_write:
    - url: http://10.1.250.232:9090/api/v1/write  # Comment same as previous case
    - url: http://10.1.250.233:9090/api/v1/write
```

metadata.yaml
```diff
@@ -27,7 +27,7 @@ resources:
     description: OCI image for Grafana Agent
 
 requires:
-  prometheus-remote-write:
+  receive-remote-write:
```
mmanciop, line 30: Should be `send-remote-write`

mmanciop:
In local testing, the Grafana Agent built from this PR is failing to push downstream to Prometheus the alert rules from the Cassandra charm.

src/charm.py
```diff
@@ -31,6 +28,10 @@
 logger = logging.getLogger(__name__)
 
 CONFIG_PATH = "/etc/agent/agent.yaml"
+METRICS_RULES_SRC_PATH = "./src/prometheus_alert_rules"
+METRICS_RULES_PATH = "./prometheus_alert_rules"
```
sed-i, line 32: We have two different paths for backwards compat?

src/charm.py
```diff
@@ -54,14 +55,23 @@ def __init__(self, *args):
...
         )
-        self._remote_write = PrometheusRemoteWriteConsumer(self)
+
+        if not os.path.isdir(self._metrics_rules_path):
+            shutil.copytree(
+                self._metrics_rules_src_path, self._metrics_rules_path, dirs_exist_ok=True
+            )
+        self._remote_write = PrometheusRemoteWriteConsumer(
+            self, alert_rules_path=self._metrics_rules_path
+        )
```
sed-i, line 74: `add_path` can be called multiple times
https://github.com/canonical/grafana-agent-k8s-operator/blob/a8f38b10fbbe5893d1c954ae22964af8fab65edf/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py#L503

so maybe the constructor should take a list of paths instead of copytree?

sed-i: The (potentially custom) relation name should be passed in as well?

src/charm.py
```diff
@@ -54,14 +55,23 @@ def __init__(self, *args):
...
         self._scrape = MetricsEndpointConsumer(self)
```
sed-i, line 75: The (potentially custom) relation name should be passed in as well?

src/charm.py
dstathis: One path is the one where alert rules for this charm are defined. The other is where we put rules to send to prometheus. Those rules can come from this charm or one related to us.

src/charm.py
dstathis: It isn't custom here though

src/charm.py
dstathis: The constructor does not support a list so while it would be a feature we could consider, it would not work here.

dstathis:
> For debugging purposes, we should annotate the Grafana Agent config with YAML comments how the IPs of the prometheus instances we remote-write to maps to their Juju identity:
> 
> ```
> integrations:
>   prometheus_remote_write:
>   - url: http://10.1.250.232:9090/api/v1/write  # Juju model X;Juju application Y; Juju unit Z
>   - url: http://10.1.250.233:9090/api/v1/write # Juju model X;Juju application Y; Juju unit W
> prometheus:
>   configs:
>   - name: agent_scraper
>     remote_write:
>     - url: http://10.1.250.232:9090/api/v1/write  # Comment same as previous case
>     - url: http://10.1.250.233:9090/api/v1/write
> ```

I think this is a bit out of scope given how much we still need to do for GA. I opened an [issue](https://github.com/canonical/grafana-agent-k8s-operator/issues/41) to add it in the future.

src/charm.py
sed-i: Perhaps add some comments about this? Currently it isn't kind to our future selves.

src/charm.py
```diff
@@ -83,6 +103,20 @@ def __init__(self, *args):
             self._on_loki_push_api_endpoint_departed,
         )
 
+    def update_metrics_rules(self, _):
+        """Copy alert rules from relations and save them to disk."""
```
sed-i, line 107: Could you please add some more context?
I am having difficulty understanding what/why this function is for.

sed-i (commented):
Could you please add a desc to the PR explaining what it is for?
And ideally add some tests to reflect that.

src/charm.py
rbarry82: I don't personally find the code that hard to follow (at least not hard enough for comments), but making it `METRICS_RULES_DEST_PATH` would pair up well.

src/charm.py
rbarry82: It deletes all existing rules, gets new ones from the relation, writes them out to files, and triggers a reload.  It could probably be named `regenerate_metrics_rules` for accuracy.

The actual method is kind of tortured, though. It would be nice to see exception handling for `rmtree()`, even if it's unlikely.

Take it or leave it, but I'd rewrite the block trying to use "native" Pathlib stuff for consistency. It ends up a lot shorter, which isn't that important, but I'd find myself spending less time wondering "are we going to use this or that variable later".

```python
for topology_identifier, rule in rules.items():
  file = Path(self._metrics_rules_path, "juju_{}_rules".format(topology_identifier))
  file.write(yaml.dump(rule))
  logger.debug(...)
```

src/charm.py
```diff
@@ -196,11 +230,7 @@ def _config_file(self, event: EventBase) -> str:
         config.update(self._server_config())
         config.update(self._integrations_config())
         config.update(self._prometheus_config())
-
-        # Don't accidentally destroy the Loki config by passing it
-        # `None` or `PebbleReady` or a `RelationEvent`
-        if isinstance(event, (LokiPushApiEndpointJoined, LokiPushApiEndpointDeparted)):
-            config.update(self._loki_config(event))
```
rbarry82, line 203: This worries me, since I don't think we have integration tests around this. `if self.model.relations["logging-provider"]: ...` below is probably safe enough, but an integration test (or even unit test which fires off multiple events) would be great.

rbarry82 (dismissed):

src/charm.py
```diff
@@ -31,6 +28,10 @@
 logger = logging.getLogger(__name__)
 
 CONFIG_PATH = "/etc/agent/agent.yaml"
+METRICS_RULES_SRC_PATH = "./src/prometheus_alert_rules"
+METRICS_RULES_PATH = "./prometheus_alert_rules"
+REMOTE_WRITE_RELATION_NAME = "send-remote-write"
+SCRAPE_RELATION_NAME = "metrics-endpoint"
```
mmanciop, line 34: We actually could import https://github.com/canonical/prometheus-k8s-operator/blob/e964e639274efff359de78b65c261c9a9cf981be/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L362

src/charm.py
```diff
@@ -71,7 +81,17 @@ def __init__(self, *args):
...
         self.framework.observe(self.on.agent_pebble_ready, self.on_pebble_ready)
         self.framework.observe(
-            self.on["prometheus-remote-write"].relation_changed, self.on_remote_write_changed
+            self.on[REMOTE_WRITE_RELATION_NAME].relation_changed, self.on_remote_write_changed
+        )
+        self.framework.observe(
+            self.on[REMOTE_WRITE_RELATION_NAME].relation_joined, self.update_metrics_rules
+        )
+        self.framework.observe(self.on.upgrade_charm, self.update_metrics_rules)
+        self.framework.observe(
```
mmanciop, line 90: Why not using the https://github.com/canonical/prometheus-k8s-operator/blob/e964e639274efff359de78b65c261c9a9cf981be/lib/charms/prometheus_k8s/v0/prometheus_scrape.py#L215 event?

src/charm.py
dstathis: I updated the method to use pathlib.

src/charm.py
dstathis: We definitely won't destroy the loki config because the method is now stateless. I can add some tests for this, but I'm not sure exactly what you are asking for. I'm sure our integration tests already fire some events.

src/charm.py
dstathis: This way feels slightly safer and also easier to read without having to open up prometheus_scrape code.

dstathis:
A note on integration tests. We definitely want some tests that do prometheus->grafana-agent->workload. I am just waiting to get a few cassandra changes in. I created an issue to track creating these tests.

mmanciop:
@dstathis https://github.com/canonical/grafana-agent-k8s-operator/runs/5763069717?check_suite_focus=true#step:4:538 speaks to a bug in the logging charm lib?

dstathis:
> @dstathis https://github.com/canonical/grafana-agent-k8s-operator/runs/5763069717?check_suite_focus=true#step:4:538 speaks to a bug in the logging charm lib?

Possibly. I'm looking into it. It looks like it's somehow running different code. I'll have to run manually to see what's going on.

sed-i:
> > @dstathis https://github.com/canonical/grafana-agent-k8s-operator/runs/5763069717?check_suite_focus=true#step:4:538 speaks to a bug in the logging charm lib?
> 
> Possibly. I'm looking into it. It looks like it's somehow running different code. I'll have to run manually to see what's going on.

Looks to me that "logging relation changed" can be called before pebble ready, and it does not have a can_connect guard.

src/charm.py
sed-i: I think what @mmanciop suggests is already practiced elsewhere: the lib is the one responsible for the relation, so if we have to hardcode that literal somewhere (in addition to metadata.yaml), it is probably better to hardcode it only once.

mmanciop:
> > > @dstathis https://github.com/canonical/grafana-agent-k8s-operator/runs/5763069717?check_suite_focus=true#step:4:538 speaks to a bug in the logging charm lib?
> > 
> > 
> > Possibly. I'm looking into it. It looks like it's somehow running different code. I'll have to run manually to see what's going on.
> 
> Looks to me that "logging relation changed" can be called before pebble ready, and it does not have a can_connect guard.

It most certainly can be called before `pebble_ready`. @Abuelodelanada to the rescue ;-)

src/charm.py
sed-i: Thanks @rbarry82.
@dstathis I think it would be nice to include in the docstring an answer to the "why" as well.

src/charm.py
```diff
@@ -54,14 +56,23 @@ def __init__(self, *args):
         super().__init__(*args)
         self._container = self.unit.get_container(self._name)
         self._stored.set_default(agent_config="")
+        self._metrics_rules_src_path = os.path.join(self.charm_dir, METRICS_RULES_SRC_PATH)
+        self._metrics_rules_path = os.path.join(self.charm_dir, METRICS_RULES_PATH)
```
sed-i, line 60: Consider adding:
- utest: forwarded rules remain intact
- itest: forwarded rule files show up when queried for via http api

Abuelodelanada:
> > > > @dstathis https://github.com/canonical/grafana-agent-k8s-operator/runs/5763069717?check_suite_focus=true#step:4:538 speaks to a bug in the logging charm lib?
> > > 
> > > 
> > > Possibly. I'm looking into it. It looks like it's somehow running different code. I'll have to run manually to see what's going on.
> > 
> > 
> > Looks to me that "logging relation changed" can be called before pebble ready, and it does not have a can_connect guard.
> 
> It most certainly can be called before `pebble_ready`. @Abuelodelanada to the rescue ;-)

Following this in https://github.com/canonical/loki-k8s-operator/issues/128

src/charm.py
rbarry82: I kind of disagree there. If we think of docstrings as primarily "references seen by developers either when editing the code or when importing it", the "why" is less important here.

From a `README` point of view for "how does this charm work?", sure. It's not necessarily important to know "why" they need to be written to disk to understand what the method is supposed to do, though, only that it's required.

src/charm.py
rbarry82: This necessarily adds coupling between an external source and code controlled by this charm, though, and `fetch-lib` could cause an unintended divergence between what's in `metadata.yaml` and what's in the code.

Yes, tests would probably catch it, but "our relation name changed" should probably be something which is intentional rather than a side effect.

src/charm.py
rbarry82: I agree that the original cause of the bug no longer exists and that the code is probably exercised already in integration tests anyway. 

My real line of thinking is "if the truthiness of `self.model.relations[key]:` changes, what happens?" And the answer to that is probably "a loki config is generated which doesn't do anything".

`integration/test_alert_rules` looks like it already covers this case, or, if it doesn't, it's very close. I certainly wouldn't expect a completely separate test for a null hypothesis with the entire code flow changed, more of "hey, if you don't think this is already covered as a side effect of something else, adding a couple of lines to break the relation at the end to fire some other relation event and make sure everything is still ok is maybe a good safety measure."

Totally up to you, though.

rbarry82 (dismissed):

dstathis:
>This necessarily adds coupling between an external source and code controlled by this charm, though, and fetch-lib could >cause an unintended divergence between what's in metadata.yaml and what's in the code.
>
>Yes, tests would probably catch it, but "our relation name changed" should probably be something which is intentional rather >than a side effect.

I don't really understand. None of those values are determined by the prometheus lib. The relation just match what is in metadata.yaml while the paths are more or less arbitrary (though they follow a standard).

src/charm.py
simskij: > The constructor does not support a list so while it would be a feature we could consider, it would not work here.

I would also argue that we in that case would add a new function called `add_paths`, which would call `add_path` once per entry.

src/charm.py
simskij: > I kind of disagree there. If we think of docstrings as primarily "references seen by developers either when editing the code or when importing it", the "why" is less important here.
> 
> From a `README` point of view for "how does this charm work?", sure. It's not necessarily important to know "why" they need to be written to disk to understand what the method is supposed to do, though, only that it's required.

The why might also change, or expand, during the lifetime of the piece of code

src/charm.py
simskij: Yes, please.

rbarry82:
> I don't really understand. None of those values are determined by the prometheus lib. The relation just match what is in metadata.yaml while the paths are more or less arbitrary (though they follow a standard).

Importing a constant from the Prometheus library instead of having it set here could cause a  `fetch-lib` to change the name to something else which _does not_ match what is in `metadata.yaml`. Then, the charm code would be watching for an event which does not exist because they are created by OF at runtime [from `metadata.yaml`, not Juju](https://github.com/canonical/operator/blob/ff30d2f4bf0d8e2f24cd11d99e3de67d8e81e3b0/ops/charm.py#L675).

In other words, I agree with the way it is written now, and think that importing from libraries is an antipattern.

mmanciop:
> > I don't really understand. None of those values are determined by the prometheus lib. The relation just match what is in metadata.yaml while the paths are more or less arbitrary (though they follow a standard).
> 
> Importing a constant from the Prometheus library instead of having it set here could cause a `fetch-lib` to change the name to something else which _does not_ match what is in `metadata.yaml`. Then, the charm code would be watching for an event which does not exist because they are created by OF at runtime [from `metadata.yaml`, not Juju](https://github.com/canonical/operator/blob/ff30d2f4bf0d8e2f24cd11d99e3de67d8e81e3b0/ops/charm.py#L675).
> 
> In other words, I agree with the way it is written now, and think that importing from libraries is an antipattern.

If we change the library name, we break `fetch-lib` already. I am failing to see why, in this scenario, is the `import` of a constant the _actual_ issue.

mmanciop:
The Grafana Agent pod is failing on my machine due to issues with the healthchecks:

```
Events:
  Type     Reason            Age                    From               Message
  ----     ------            ----                   ----               -------
  Warning  FailedScheduling  4m33s                  default-scheduler  0/1 nodes are available: 1 pod has unbound immediate PersistentVolumeClaims.
  Normal   Scheduled         4m31s                  default-scheduler  Successfully assigned cos/grafana-agent-k8s-0 to boombox
  Normal   Pulled            4m31s                  kubelet            Container image "jujusolutions/jujud-operator:2.9.27" already present on machine
  Normal   Created           4m31s                  kubelet            Created container charm-init
  Normal   Started           4m31s                  kubelet            Started container charm-init
  Normal   Pulling           4m29s                  kubelet            Pulling image "grafana/agent:v0.20.1"
  Normal   Pulled            3m40s                  kubelet            Successfully pulled image "grafana/agent:v0.20.1" in 48.821275418s
  Warning  Unhealthy         3m21s (x2 over 3m31s)  kubelet            Startup probe failed: Get "http://10.1.151.76:3856/startup": dial tcp 10.1.151.76:3856: connect: network is unreachable
  Normal   Killing           3m21s                  kubelet            Container charm failed startup probe, will be restarted
  Warning  Unhealthy         3m6s                   kubelet            Liveness probe failed: Get "http://10.1.151.76:38813/v1/health?level=alive": dial tcp 10.1.151.76:38813: connect: network is unreachable
  Normal   Pulled            2m51s (x2 over 4m29s)  kubelet            Container image "jujusolutions/charm-base:ubuntu-20.04" already present on machine
  Normal   Created           2m51s (x2 over 4m29s)  kubelet            Created container charm
  Normal   Started           2m51s (x2 over 4m29s)  kubelet            Started container charm
  Normal   Killing           2m51s                  kubelet            Container agent failed liveness probe, will be restarted
  Normal   Pulled            2m51s                  kubelet            Container image "grafana/agent:v0.20.1" already present on machine
  Warning  Unhealthy         2m51s (x4 over 3m6s)   kubelet            Readiness probe failed: Get "http://10.1.151.76:38813/v1/health?level=ready": dial tcp 10.1.151.76:38813: connect: network is unreachable
  Normal   Created           2m51s (x2 over 3m40s)  kubelet            Created container agent
  Normal   Started           2m51s (x2 over 3m40s)  kubelet            Started container agent
```

rbarry82:
> > > I don't really understand. None of those values are determined by the prometheus lib. The relation just match what is in metadata.yaml while the paths are more or less arbitrary (though they follow a standard).
> > 
> > 
> > Importing a constant from the Prometheus library instead of having it set here could cause a `fetch-lib` to change the name to something else which _does not_ match what is in `metadata.yaml`. Then, the charm code would be watching for an event which does not exist because they are created by OF at runtime [from `metadata.yaml`, not Juju](https://github.com/canonical/operator/blob/ff30d2f4bf0d8e2f24cd11d99e3de67d8e81e3b0/ops/charm.py#L675).
> > In other words, I agree with the way it is written now, and think that importing from libraries is an antipattern.
> 
> If we change the library name, we break `fetch-lib` already. I am failing to see why, in this scenario, is the `import` of a constant the _actual_ issue.

Not the library name -- I am saying: if we import a constant for the relation name _from_ the library, and it changes there, it will be out of sync with what is in `metadata.yaml` (and as a consequence, the properties added to the charm for  `foo_relation_...`) as a "side-effect" of `fetch-lib`.

We'd need to update either way, but from a process POV, `self.framework.observe(self.on[FOO]...)` seems, to me, better suited as an intentional change, rather than "I ran `fetch-lib` and now my integration tests fail because my imported constant changed and it doesn't match `metadata.yaml` anymore"

Just one person's opinion. Sure, we _can_ import it from libs, but should we? I vote :-1:

mmanciop:
> We'd need to update either way, but from a process POV, `self.framework.observe(self.on[FOO]...)` seems, to me, better suited as an intentional change, rather than "I ran `fetch-lib` and now my integration tests fail because my imported constant changed and it doesn't match `metadata.yaml` anymore"
> 
> Just one person's opinion. Sure, we _can_ import it from libs, but should we? I vote 👎

My (hopefully unsurprising) PoV is that public constants in a charm library are API and we need to put the same care in keeping them stable as we do with methods and classes.

src/charm.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, *args):
 
         self.framework.observe(self.on.agent_pebble_ready, self.on_pebble_ready)
         self.framework.observe(
-            self.on[REMOTE_WRITE_RELATION_NAME].relation_changed, self.on_remote_write_changed
+            self._remote_write.on.endpoints_changed, self.on_remote_write_changed
```
mmanciop, line 85: Isn’t it likely we have the _same_ bug on line 88? And a similar one on 92/95?

Also: where are the tests? ;-)

mmanciop:
Itests still borked https://github.com/canonical/grafana-agent-k8s-operator/runs/5860652469?check_suite_focus=true#step:4:563

mmanciop:
> With the Grafana Agent from this PR, I do not manage to get metrics scraped by Grafana Agent into Prometheus via remote write.
> 
> This setup shows `traefik-k8s` alert rules in `prometheus`, but no metrics whatsoever:
> 
> ```
> Model  Controller   Cloud/Region        Version  SLA          Timestamp
> cos    development  microk8s/localhost  2.9.27   unsupported  15:00:15+02:00
> 
> App                           Version  Status  Scale  Charm                         Channel  Rev  Address         Exposed  Message
> grafana-agent-k8s                      active      1  grafana-agent-k8s                        1  10.152.183.192  no
> prometheus                             active      1  prometheus-k8s                edge      29  10.152.183.103  no
> traefik-k8s                            active      1  traefik-k8s                   edge      54  10.152.183.230  no
> 
> Unit                             Workload  Agent  Address      Ports  Message
> grafana-agent-k8s/0*             active    idle   10.1.151.77
> prometheus/0*                    active    idle   10.1.151.75
> traefik-k8s/0*                   active    idle   10.1.151.78
> 
> Relation provider                              Requirer                                               Interface                Type     Message
> prometheus:metrics-endpoint  grafana-agent-k8s:metrics-endpoint                     prometheus_scrape        regular
> prometheus:prometheus-peers                   prometheus:prometheus-peers                           prometheus_peers         peer
> prometheus:receive-remote-write                grafana-agent-k8s:send-remote-write                    prometheus_remote_write  regular
> traefik-k8s:metrics-endpoint                   grafana-agent-k8s:metrics-endpoint  prometheus_scrape        regular
> ```

I found the issue and it blows my mind: the `agent` integration to do remote write is mutually exclusive with the prometheus remote write. That is, if you tell Grafana Agent to sent its own metrics upstream with an `agent -> integrations` config, it ignores the remote_write you set up as a `prometheus -> config`.

src/charm.py
mmanciop: Fixed in 621103cb14fa759dce2ee87bca28e1274241c3ae

dstathis:
@mmanciop Why did you change the agent config to not use the integration? It works just fine. The only issue was the restart/replan thing. Was there another reason?

tests/unit/test_charm.py
```diff
@@ -81,29 +68,27 @@ def read(self, *args, **kwargs):
...
+@patch.object(Container, "restart", new=lambda x, y: True)
 class TestCharm(unittest.TestCase):
     @patch("charm.KubernetesServicePatch", lambda x, y: None)
     def setUp(self):
         self.harness = Harness(GrafanaAgentOperatorCharm)
         self.addCleanup(self.harness.cleanup)
         self.harness.set_model_info(name="lma", uuid="1234567890")
         self.harness.set_leader(True)
-        self.harness.begin()
+        self.harness.begin_with_initial_hooks()
```
sed-i, line 79: [Apparently](https://github.com/canonical/operator/pull/727#issuecomment-1079096588) we need to call `pebble_ready` ourselves after `begin_with_initial_hooks`.
Since there is an `except (FileNotFoundError, PathError)` above, it would be nice to test the outcome before and after calling `pebble_ready`.

sed-i (commented):
Re itest failing: I see similar errors too [elsewhere](https://github.com/canonical/prometheus-k8s-operator/runs/5996773670?check_suite_focus=true).

tests/unit/test_charm.py
```diff
@@ -81,29 +68,27 @@ def read(self, *args, **kwargs):
...
         responses.add(
             responses.POST,
             "http://localhost/-/reload",
             status=200,
         )
 
-        rel_id = self.harness.add_relation("prometheus-remote-write", "prometheus")
+        agent_container = self.harness.charm.unit.get_container("agent")
+
+        rel_id = self.harness.add_relation("send-remote-write", "prometheus")
```
sed-i, line 91: This should be called before the first set_leader:
https://github.com/canonical/operator/pull/733

CONTRIBUTING.md
```diff
@@ -61,7 +61,7 @@ The core implementation of this charm is represented by the [`GrafanaAgentOperat
 
 - `self.on.install`: In this event we patch K8s service ports.
```
balbirthomas, line 62: The PR description is rather terse : "Clean up the implementation of Prometheus Remote Write in Grafana Agent." . It is not clear why the clean up was required and what were the cleanups that were required.

dstathis: The most important reason was that alerts were not forwarded to prometheus

tests/unit/test_charm.py
dstathis: That is for peer relations. This is a regular relation.

tests/unit/test_charm.py
dstathis: Resolving because itests are passing

sed-i: Do you mind rerunning the tests with the following added after `begin_with_initial_hooks`?
`self.harness.container_pebble_ready("agent")`

src/charm.py
```diff
@@ -150,31 +173,25 @@ def _update_config(self, event=None):
             # Pebble is not ready yet so no need to update config
             self.unit.status = WaitingStatus("waiting for agent container to start")
             return
+            return
```
sed-i, line 176: Remove duplicated?

sed-i (approved):

tests/unit/test_charm.py
```diff
@@ -119,8 +104,6 @@ def test_remote_write_configuration(self, mock_push: MagicMock, mock_restart: Ma
             {"remote_write": json.dumps({"url": "http://1.1.1.2:9090/api/v1/write"})},
         )
 
-        path, content = mock_push.call_args[0]
-        content = yaml.safe_load(content)
         expected_config: Dict[str, Any] = {
             "integrations": {
```
mmanciop, line 108: I think it might make sense to use a rewrite_rule to change the `job_name` label, which currently has the `integrations/agent` value, to be in line with those we use elsewhere based on topology:
Screenshot from 2022-04-28 18-14-16

mmanciop (commented):
Alert rule propagation does not seem to work in the following scenario:

```sh
juju add-model cos
juju deploy cos-lite --channel=edge --trust
juju offer prometheus:receive-remote-write prw
juju add-model agent
juju deploy ./*.charm grafana-agent --resource agent-image='grafana/agent:v0.20.1'
juju deploy spring-music --channel=edge
juju relate spring-music:metrics-endpoint grafana-agent
juju consume cos.prw prometheus
juju relate grafana-agent prometheus
```

Then open the Prometheus ui, go to the alert rules, despair: metrics from Spring Music come in, alert rules do not.

mmanciop:
> > Alert rule propagation does not seem to work in the following scenario:
> > 
> > 1. juju add-model cos
> > 2. juju deploy cos-lite --channel=edge --trust
> > 3. juju offer prometheus:send-remote-write prw
> > 4. juju add-model agent
> > 5. juju deploy ./*.charm grafana-agent ...
> > 6. juju deploy spring-music --channel=edge
> > 7. juju relate spring-music:metrics-endpoint grafana-agent
> > 8. juju consume cos.prw prometheus
> > 9. juju relate grafana-agent prometheus
> > 10. Open the Prometheus ui, go to the alert rules, despair
> > 
> > Metrics from Spring Music come in, alert rules do not
> 
> @mmanciop I tried this and it's working just fine for me. Could you possibly do it again and copy paste the exact commands?

Wanna bet it is a race condition? ;-)

tests/unit/test_charm.py
mmanciop: Addressed in 073a3afdc44640a180fca43dde95178a44041dcd

src/charm.py
dstathis: Done.

tests/unit/test_charm.py
```diff
@@ -81,29 +68,27 @@ def read(self, *args, **kwargs):
 ]
 
 
+@patch.object(Container, "restart", new=lambda x, y: True)
```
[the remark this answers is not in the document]
dstathis: The tests break if I remove it.

tests/unit/test_charm.py
dstathis: The tests pass either way. Should I add the line permanently?

CONTRIBUTING.md
```diff
@@ -61,7 +61,7 @@ The core implementation of this charm is represented by the [`GrafanaAgentOperat
...
 - `self.on.agent_pebble_ready`: In this event the charm builds the Pebble layer to be added in the workload charm. This Pebble layer will manage the execution of Promtail binary
-- `self.on["prometheus-remote-write"].relation_changed`: In this event the Grafana agent config is updated with remote-write settings.
+- `self.on["send-remote-write"].relation_changed`: In this event the Grafana agent config is updated with remote-write settings.
```
rbarry82, line 64:
```suggestion
- `self.on.send_remote_write_relation_changed`: In this event the Grafana agent config is updated with remote-write settings.
```

(This isn't actually important unless you care about it for taste, but I think accepting the suggestion also keeps the +1)

src/charm.py
```diff
@@ -91,19 +120,13 @@ def _on_loki_push_api_endpoint_departed(self, event) -> None:
...
             event: The event object of the pebble ready event
         """
-        container = event.workload
-        if not self._stored.agent_config:
-            config = self._config_file(event)
-            container.push(CONFIG_PATH, config)
-            self._stored.agent_config = config
-        else:
-            container.push(CONFIG_PATH, self._stored.agent_config)
+        self._container.push(CONFIG_PATH, yaml.dump(self._config_file()), make_dirs=True)
```
rbarry82, line 129: If we get another PR, a guard wouldn't hurt here. The container is _almost certain_ to be contacted right after `pebble_ready`, but better to be safe...

src/charm.py
```diff
@@ -83,6 +99,19 @@ def __init__(self, *args):
...
+    def update_alerts_rules(self, _):
+        """Copy alert rules from relations and save them to disk."""
+        rules = self._scrape.alerts()
+        shutil.rmtree(self._metrics_rules_dest_path)
+        shutil.copytree(self._metrics_rules_src_path, self._metrics_rules_dest_path)
+        for topology_identifier, rule in rules.items():
+            file_handle = pathlib.Path(
+                self._metrics_rules_dest_path, "juju_{}.rules".format(topology_identifier)
+            )
+            file_handle.write_text(yaml.dump(rule))
```
rbarry82, line 111: If we get another push, this wouldn't hurt to change. The name is ok, but it makes me feel like it's a handle which should be closed so it doesn't leak rather than `Path` hiding it.

src/charm.py
```diff
@@ -223,14 +235,21 @@ def _integrations_config(self) -> dict:
         juju_application = self.model.app.name
         juju_unit = self.unit.name
 
+        job_name = f"juju_{juju_model}_{juju_model_uuid}_{juju_application}_self-monitoring"
```
rbarry82, line 238: Safer to avoid f-strings just in case this ends up as a machine charm somewhere. It's not _really_ k8s specific, and could wind up on python 3.5 or something.

rbarry82 (approved):
Tests are solid, love seeing the mocking gone, rewrite rule is nice.

Feel free to just resolve any of my comments if you don't expect another push. They're not critical in any way.

balbirthomas (approved):

Abuelodelanada (approved):

src/charm.py
dstathis: This is not in a library so it will never end up on python 3.5.

balbirthomas: Ack. Python 3.5 was only mandatory for libraries. I meant for charms it is ok to use upto 3.8 if I am not mistaken.

PR #21, opened 2021-11-16
Add integration test

Abuelodelanada:


sed-i (approved):
Adding another test that confirms relating to relevant charms results in active/idle should be easy enough as well. I would split each relation to a different test. [Example](https://github.com/sed-i/lma-rules-operator/blob/a5213256dfb23d2ffe8b726921e44037b09d610c/tests/integration/test_charm.py#L89)

tests/integration/test_charm.py
```diff
@@ -3,14 +3,34 @@
...
+    # due to a juju bug, occasionally some charms finish a startup sequence with "waiting for IP
+    # address"
+    # issuing dummy update_status just to trigger an event
+    await ops_test.model.set_config({"update-status-hook-interval": "10s"})
+
+    await ops_test.model.wait_for_idle(apps=["agent"], status="active", timeout=1000)
+    assert ops_test.model.applications["agent"].units[0].workload_status == "active"
+
+    # effectively disable the update status from firing
+    await ops_test.model.set_config({"update-status-hook-interval": "60m"})
```
dstathis, line 36: why?

dstathis (approved):

Abuelodelanada:
> Adding another test that confirms relating to relevant charms results in active/idle should be easy enough as well. I would split each relation to a different test. [Example](https://github.com/sed-i/lma-rules-operator/blob/a5213256dfb23d2ffe8b726921e44037b09d610c/tests/integration/test_charm.py#L89)

Yes, I was reading the same example! Thanks!

.github/workflows/run_tests.yaml
```diff
@@ -35,3 +35,17 @@ jobs:
...
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Setup operator environment
+      uses: charmed-kubernetes/actions-operator@main
+      with:
+        provider: microk8s
+    - name: Run Grafana-agent tests
+      run: tox -e integration
+    - name: Run lma bundle tests
+      run: tox -e integration-lma
```
sed-i, line 51: Is this charm (grafana-agent-operator) going to have any impact on the lma bundle tests? i.e. why run `integration-lma` here?

tests/integration/test_charm.py
```diff
@@ -3,14 +3,40 @@
...
+
+
+async def test_relating_to_loki(ops_test):
+    await ops_test.model.deploy("loki-k8s", channel="edge", application_name="loki")
+    await ops_test.model.add_relation("loki", "agent")
+    await ops_test.model.wait_for_idle(apps=["loki"], status="active", timeout=1000)
```
sed-i, line 42: Both loki and agent should be active I believe
```suggestion
    await ops_test.model.wait_for_idle(apps=["loki", "agent"], status="active", timeout=1000)
```
or if you only have those two charms deployed then
```suggestion
    await ops_test.model.wait_for_idle(status="active", timeout=1000)
```

.github/workflows/run_tests.yaml
Abuelodelanada: AFAIU yes, but will remove it now, and add it later when we write the tests for that

PR #35, opened 2022-02-03
Loki config missing if VM running juju and microk8s is restarted or pod is deleted

### LINKED ISSUE #34: Loki config missing if VM running juju and microk8s is restarted.
Loki config missing in grafana-agent charm.

how can it be that with the latest Grafana Agent edge, when I relate Grafana Agent with Loki, I get no loki configurations generated?

```bash
michele@boombox:~$ juju status --relations
Model  Controller   Cloud/Region        Version  SLA          Timestamp
lma    development  microk8s/localhost  2.9.22   unsupported  18:56:09+01:00

App            Version  Status  Scale  Charm              Store     Channel  Rev  OS          Address         Message
alertmanager            active      1  alertmanager-k8s   charmhub  edge       7  kubernetes  10.152.183.187  
grafana                 active      1  grafana-k8s        charmhub  edge      14  kubernetes  10.152.183.218  
grafana-agent           active      2  grafana-agent-k8s  charmhub  edge       4  kubernetes  10.152.183.232  
loki                    active      1  loki-k8s           charmhub  edge      11  kubernetes  10.152.183.168  
prometheus              active      1  prometheus-k8s     charmhub  edge      13  kubernetes  10.152.183.139  

Unit              Workload  Agent  Address      Ports  Message
alertmanager/0*   active    idle   10.1.151.76         
grafana-agent/0*  active    idle   10.1.151.80         
grafana-agent/1   active    idle   10.1.151.81         
grafana/0*        active    idle   10.1.151.77         
loki/0*           active    idle   10.1.151.78         
prometheus/0*     active    idle   10.1.151.79         

Relation provider          Requirer                 Interface              Type     Message
alertmanager:alerting      prometheus:alertmanager  alertmanager_dispatch  regular  
alertmanager:replicas      alertmanager:replicas    alertmanager_replica   peer     
grafana:grafana-peers      grafana:grafana-peers    grafana_peers          peer     
loki:grafana-source        grafana:grafana-source   grafana_datasource     regular  
loki:logging               grafana-agent:logging    loki_push_api          regular  
prometheus:grafana-source  grafana:grafana-source   grafana_datasource     regular 
```
```bash
michele@boombox:~$ microk8s.kubectl exec -it grafana-agent-0 -n lma -c agent -- cat /etc/agent/agent.yaml
integrations:
  agent:
    enabled: true
    relabel_configs:
    - regex: (.*)
      replacement: lma_f149bca2-4d3e-4289-8b72-6b66b40d14dc_grafana-agent_grafana-agent/0
      target_label: instance
    - replacement: grafana-agent-k8s
      source_labels:
      - __address__
      target_label: juju_charm
    - replacement: lma
      source_labels:
      - __address__
      target_label: juju_model
    - replacement: f149bca2-4d3e-4289-8b72-6b66b40d14dc
      source_labels:
      - __address__
      target_label: juju_model_uuid
    - replacement: grafana-agent
      source_labels:
      - __address__
      target_label: juju_application
    - replacement: grafana-agent/0
      source_labels:
      - __address__
      target_label: juju_unit
  prometheus_remote_write: []
loki: {}
prometheus:
  configs:
  - name: agent_scraper
    remote_write: []
    scrape_configs: []
server:
  log_level: info
```

Abuelodelanada:
This PR fixes #34.

There are 2 way to test this PR:

#### Restart VM

- Deploy Loki and deploy 2 Grafana agents
- Relate both charms
- Check the grafana agent config file has the `loki` section
- **Stop the multipass VM** in which Juju and microk8s are running.
- Start the multipass VM again.
- Check that once microk8s is up and running again the grafana agent config has the `loki` section

#### Delete POD

- Deploy Loki and deploy 2 Grafana agents
- Relate both charms
- Check the grafana agent config file has the `loki` section
- **Delete** the grafana agent pod with: `microk8s.kubectl delete pods grafana-agent-k8s-0 -n observability`
- Wait until the pod is deleted and Juju creates a new unit to replace the deleted one.
- Check grafana agent config has the `loki` section

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -168,6 +168,11 @@ class GrafanaAgentOperatorCharm(CharmBase):
                 self._container.restart(self._name)
                 self.unit.status = ActiveStatus()
         except APIError as e:
+            # When Juju creates a new unit (because the previous one was killed)
+            # the `_on_loki_push_api_endpoint_joined` event is fired before `pebble-ready` event,
+            # BUT when Pebble is actually ready (self._container.can_connect() is True).
+            # APIError is raised because "agent" service doesn't exist yet since we add that
+            # layer in on_pebble_ready.
             self.unit.status = WaitingStatus(str(e))
             event.defer()
         except GrafanaAgentReloadError as e:
--- a/src/charm.py
+++ b/src/charm.py
@@ -53,7 +53,7 @@ class GrafanaAgentOperatorCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
         self._container = self.unit.get_container(self._name)
-        self._stored.set_default(k8s_service_patched=False, config="")
+        self._stored.set_default(agent_config="")
         self.service_patch = KubernetesServicePatch(
             self,
             [
@@ -98,8 +98,13 @@ class GrafanaAgentOperatorCharm(CharmBase):
             event: The event object of the pebble ready event
         """
         container = event.workload
-        config = self._config_file(event)
-        container.push(CONFIG_PATH, config)
+        if not self._stored.agent_config:
+            config = self._config_file(event)
+            container.push(CONFIG_PATH, config)
+            self._stored.agent_config = config
+        else:
+            container.push(CONFIG_PATH, self._stored.agent_config)
+
         pebble_layer = {
             "summary": "agent layer",
             "description": "pebble config layer for Grafana Agent",
@@ -157,6 +162,7 @@ class GrafanaAgentOperatorCharm(CharmBase):
         try:
             if yaml.safe_load(config) != yaml.safe_load(old_config):
                 self._container.push(CONFIG_PATH, config)
+                self._stored.agent_config = config
                 # FIXME: #19
                 # self._reload_config()
                 self._container.restart(self._name)
--- a/src/charm.py
+++ b/src/charm.py
@@ -149,7 +149,6 @@ class GrafanaAgentOperatorCharm(CharmBase):
         if not self._container.can_connect():
             # Pebble is not ready yet so no need to update config
             self.unit.status = WaitingStatus("waiting for agent container to start")
-            event.defer()
             return
 
         config = self._config_file(event)
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,7 @@ from ops.charm import CharmBase, RelationChangedEvent
 from ops.framework import EventBase, StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
-from ops.pebble import PathError
+from ops.pebble import APIError, PathError
 from requests import Session
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
@@ -122,14 +122,14 @@ class GrafanaAgentOperatorCharm(CharmBase):
 
         self._update_status()
 
-    def on_scrape_targets_changed(self, _) -> None:
+    def on_scrape_targets_changed(self, event) -> None:
         """Event handler for the scrape targets changed event."""
-        self._update_config()
+        self._update_config(event)
         self._update_status()
 
-    def on_remote_write_changed(self, _: RelationChangedEvent) -> None:
+    def on_remote_write_changed(self, event: RelationChangedEvent) -> None:
         """Event handler for the remote write changed event."""
-        self._update_config()
+        self._update_config(event)
         self._update_status()
 
     def _update_status(self) -> None:
@@ -149,6 +149,7 @@ class GrafanaAgentOperatorCharm(CharmBase):
         if not self._container.can_connect():
             # Pebble is not ready yet so no need to update config
             self.unit.status = WaitingStatus("waiting for agent container to start")
+            event.defer()
             return
 
         config = self._config_file(event)
@@ -167,6 +168,9 @@ class GrafanaAgentOperatorCharm(CharmBase):
                 # self._reload_config()
                 self._container.restart(self._name)
                 self.unit.status = ActiveStatus()
+        except APIError as e:
+            self.unit.status = WaitingStatus(str(e))
+            event.defer()
         except GrafanaAgentReloadError as e:
             self.unit.status = BlockedStatus(str(e))
 


src/charm.py
```diff
@@ -98,8 +98,13 @@ def on_pebble_ready(self, event: EventBase) -> None:
             event: The event object of the pebble ready event
         """
         container = event.workload
-        config = self._config_file(event)
-        container.push(CONFIG_PATH, config)
+        if not self._stored.agent_config:
```
sed-i, line 101: Could you please add a comment for what this `if` is checking? I don't understand why the old code is wrong.

Also, recently @mmanciop (?) encountered scenarios where stored wasn't reliable for this purpose here so now we're moving to storing in peer relation app data (if it's the same stuff for all units; you cannot use peer unit data because in the case of a single unit you won't have that bag).

src/charm.py
```diff
@@ -157,10 +163,14 @@ def _update_config(self, event=None):
...
+                self._stored.agent_config = config
                 # FIXME: #19
                 # self._reload_config()
                 self._container.restart(self._name)
                 self.unit.status = ActiveStatus()
+        except APIError as e:
+            self.unit.status = WaitingStatus(str(e))
+            event.defer()
         except GrafanaAgentReloadError as e:
             self.unit.status = BlockedStatus(str(e))
```
sed-i, line 175: Would it make sense to narrow down the scope of the try, maybe break it into two separate ones?

sed-i (commented):
I'm surprised the pod deletion scenario wasn't picked up by the itest
https://github.com/canonical/grafana-agent-k8s-operator/blob/main/tests/integration/test_kubectl_delete.py
Mind giving it a look so it captures it?

src/charm.py
```diff
@@ -144,6 +149,7 @@ def _update_config(self, event=None):
         if not self._container.can_connect():
             # Pebble is not ready yet so no need to update config
             self.unit.status = WaitingStatus("waiting for agent container to start")
+            event.defer()
```
dstathis, line 152: Why do we need to defer here? If can_connect is false, then we know we will get another call when pebble is ready.

src/charm.py
```diff
@@ -157,10 +163,14 @@ def _update_config(self, event=None):
...
             if yaml.safe_load(config) != yaml.safe_load(old_config):
                 self._container.push(CONFIG_PATH, config)
+                self._stored.agent_config = config
                 # FIXME: #19
                 # self._reload_config()
                 self._container.restart(self._name)
                 self.unit.status = ActiveStatus()
+        except APIError as e:
+            self.unit.status = WaitingStatus(str(e))
+            event.defer()
```
dstathis, line 173: What does APIError mean here? I have some doubts that deferring is the correct action.

src/charm.py
```diff
@@ -98,8 +98,13 @@ def on_pebble_ready(self, event: EventBase) -> None:
             event: The event object of the pebble ready event
         """
         container = event.workload
-        config = self._config_file(event)
-        container.push(CONFIG_PATH, config)
+        if not self._stored.agent_config:
+            config = self._config_file(event)
+            container.push(CONFIG_PATH, config)
+            self._stored.agent_config = config
```
dstathis, line 104: I don't understand why we are storing the config. We have a method that generates the config file. Why not just call that each time we need it? Is it too slow?

src/charm.py
```diff
@@ -53,7 +53,7 @@ class GrafanaAgentOperatorCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
         self._container = self.unit.get_container(self._name)
-        self._stored.set_default(k8s_service_patched=False, config="")
```
dstathis, line 56: Is `k8s_service_patched` not needed anymore?

dstathis (changes requested):

src/charm.py
Abuelodelanada: The config file is built on pebble_ready and completed once you relate grafana agent with loki.

I'm saving the configuration in stored because if you have a grafana agent and loki charms related, when you reboot the VM in which microk8s is running, the `loki_push_api_endpoint_joined` is not fired again and the config will be in a different state (without loki section)

src/charm.py
Abuelodelanada: No, now we are using the patch from the observability_libs:

https://github.com/canonical/grafana-agent-k8s-operator/pull/33

src/charm.py
sed-i: The old code,
```python
config = self._config_file(event)
container.push(CONFIG_PATH, config)
```
seems to be doing the right thing already, but I guess I'm missing something here.
Ideally the full config should be constructible upon re-entry without stored state. Would that be possible here?

rbarry82: Arguably, we may want to see whether the config is actually different, so it's not constantly interrupting the binary and the flow of logs

mmanciop: Well we are using the reload web api, I don’t think it is a big disruption

src/charm.py
Abuelodelanada: When you kill a pod, Juju will create a new one to replace it.

When creating the new unit the `_on_loki_push_api_endpoint_joined` event is fired before `pebble-ready`event, BUT when Pebble is ready (that why `self._container.can_connect()` in line 149 is `True`).

`APIError` is raised because `agent` service doesn't exist yet since we add that layer in `on_pebble_ready`:

```
  File "./src/charm.py", line 88, in _on_loki_push_api_endpoint_joined
    self._update_config(event)
  File "./src/charm.py", line 169, in _update_config
    self._container.restart(self._name)
  File "/var/lib/juju/agents/unit-grafana-agent-k8s-0/charm/venv/ops/model.py", line 1159, in restart
    self._pebble.start_services(service_names)
  File "/var/lib/juju/agents/unit-grafana-agent-k8s-0/charm/venv/ops/pebble.py", line 1251, in start_services
    return self._services_action('start', services, timeout, delay)
  File "/var/lib/juju/agents/unit-grafana-agent-k8s-0/charm/venv/ops/pebble.py", line 1304, in _services_action
    resp = self._request('POST', '/v1/services', body=body)
  File "/var/lib/juju/agents/unit-grafana-agent-k8s-0/charm/venv/ops/pebble.py", line 1115, in _request
    response = self._request_raw(method, path, query, headers, data)
  File "/var/lib/juju/agents/unit-grafana-agent-k8s-0/charm/venv/ops/pebble.py", line 1155, in _request_raw
    raise APIError(body, code, status, message)
ops.pebble.APIError: cannot start services: service "agent" does not exist
```

That's why I'm catching that exception and deferring the event.

src/charm.py
Abuelodelanada: Look at [this comment](https://github.com/canonical/grafana-agent-k8s-operator/pull/35#discussion_r799036990)

src/charm.py
Abuelodelanada: @rbarry82 before this PR, when you reboot the VM in which microk8s is running the config ends up with this:

```
...
loki: {}
prometheus:
  configs:
  - name: agent_scraper
    remote_write: []
    scrape_configs: []
server:
  log_level: info
  ```
But should be:

```
...
loki:
  configs:
  - clients:
    - url: http://loki-k8s-0.loki-k8s-endpoints.observability.svc.cluster.local:3100/loki/api/v1/push
    name: promtail
    positions:
      filename: /tmp/positions.yaml
    scrape_configs:
    - job_name: loki
      loki_push_api:
        server:
          grpc_listen_port: 3600
          http_listen_port: 3500
prometheus:
  configs:
  - name: agent_scraper
    remote_write: []
    scrape_configs: []
server:
  log_level: info
```

File: src/charm.py (hunk 3 above)
Comment (Abuelodelanada) [reply src/charm.py:152]: You are right, it's not necessary.

File: src/charm.py (hunk 1 above)
Comment (Abuelodelanada) [reply src/charm.py:101]: > Well we are using the reload web api, I don’t think it is a big disruption

@mmanciop we wanted to use the reload web api, but we [found a bug](https://github.com/canonical/grafana-agent-k8s-operator/issues/19):

```python
# FIXME: #19
# self._reload_config()
self._container.restart(self._name)
```

src/charm.py
sed-i: > When creating the new unit the `_on_loki_push_api_endpoint_joined` event is fired before `pebble-ready`event

I would put that in a comment.

src/charm.py
rbarry82: @Abuelodelanada Granted, I was making a comment on "we may not want to rebuild the config and restart it on every config change -- only if it's different", which is a good reason to either pull it from the container and compare or to keep it stored off somewhere.

Abuelodelanada: @rbarry82 we are doing this. The lines you are reading (101-104) are executed `on_pebble_ready` method that is executed only while deploying the unit.

I think we are doing what are you proposing:

```python
        config = self._config_file(event)

        try:
            old_config = self._container.pull(CONFIG_PATH)
        except PathError:
            # If the file does not yet exist, pebble_ready has not run yet
            pass

        try:
            if yaml.safe_load(config) != yaml.safe_load(old_config):
                self._container.push(CONFIG_PATH, config)
                self._stored.agent_config = config
                # FIXME: #19
                # self._reload_config()
                self._container.restart(self._name)
```

Or I'm missing something?

src/charm.py
dstathis: I think the safest way to make this work is to have a method that builds the *whole* config file and call it in both pebble_ready and relation_joined. The method can simply exit if resources are not yet ready and it will be called when the second hook fires.

The reason I think this is better is we have all the information needed to construct the config file statelessly and added a stored state introduces another moving part which could introduce bugs.

src/charm.py
Abuelodelanada: done

sed-i (approved):

PR #44, opened 2022-04-07
FIX #42 - Downscaling Loki wipes its configuration

### LINKED ISSUE #42: Downscaling Loki results in a bad config
### Bug Description

When a grafana-agent charm is related with (let's say) 3 Loki units, and Loki is down scaled to (let's say) 2 units, the `loki`section in emptied.

 

### To Reproduce

1. Deploy grafana-agent
2. Deploy 3 loki units
3. Relate both charms `juju add-relation grafana-agent loki`
4. Down scale Loki `juju scale-application loki 2`
5. Verify that `loki` section in grafana-agent config file is empty `juju ssh --container agent agent/0 cat /etc/agent/agent.yaml`

### Environment

Grafana agent: `main` branch

### Relevant log output

```shell
This is the config file of a grafana-agent charm related with 2 Loki units and then downscaled to 1 Loki unit. 

juju ssh --container agent agent/0 cat /etc/agent/agent.yaml
integrations:
  agent:
    enabled: true
    relabel_configs:
    - regex: (.*)
      replacement: cos_1651a423-4b8c-4f27-8fbd-bff940439eed_agent_agent/0
      target_label: instance
    - replacement: grafana-agent-k8s
      source_labels:
      - __address__
      target_label: juju_charm
    - replacement: cos
      source_labels:
      - __address__
      target_label: juju_model
    - replacement: 1651a423-4b8c-4f27-8fbd-bff940439eed
      source_labels:
      - __address__
      target_label: juju_model_uuid
    - replacement: agent
      source_labels:
      - __address__
      target_label: juju_application
    - replacement: agent/0
      source_labels:
      - __address__
      target_label: juju_unit
  prometheus_remote_write: []
loki: {}
prometheus:
  configs:
  - name: agent_scraper
    remote_write: []
    scrape_configs: []
server:
  log_level: info
```

### Additional context

We must fix `_loki_config` method in charm.py

Abuelodelanada:
## Issue

See #42 

## Solution

Now we are no longer checking the event type, we are only checking if we have `loki_endpoints` provided by the Provider side of the relation

## Testing Instructions

1. Deploy grafana-agent
2. Deploy 3 loki units
3. Relate both charms `juju add-relation grafana-agent loki`
4. Down scale Loki `juju scale-application loki 2`
5. Verify that `loki` section in grafana-agent config file is NOT empty: `juju ssh --container agent agent/0 cat /etc/agent/agent.yaml`


tests/unit/test_charm.py
```diff
@@ -267,6 +267,54 @@ def test__on_loki_push_api_endpoint_departed(
         self.assertEqual(path, "/etc/agent/agent.yaml")
         self.assertTrue(yaml.safe_load(content)["loki"] == {})
 
+    @responses.activate
+    @patch.object(Container, "pull", new=pull_empty_fake_file)
```
mmanciop, line 271: Container puss and pull is now supported by harness, see e.g. https://github.com/canonical/grafana-agent-k8s-operator/pull/17/commits/621103cb14fa759dce2ee87bca28e1274241c3ae

mmanciop (commented):
I’d like to see a cleanup of the excessive mocking in the tests. The rest LGTM.

tests/unit/test_charm.py
```diff
@@ -250,13 +250,13 @@ def test__on_loki_push_api_endpoint_joined(
...
     @patch.object(Container, "push")
-    def test__on_loki_push_api_endpoint_departed(
+    def test__on_loki_push_api_endpoint_departed_0_endpoints(
         self, mock_push: MagicMock, mock_restart: MagicMock
     ):
         """Test Loki config is not in config file when LokiPushApiEndpointDeparted is fired."""
         mock_restart.restart.return_value = True
         self.harness.charm._loki_consumer = Mock()
-        self.harness.charm._loki_consumer.loki_push_api = "http://loki:3100:/loki/api/v1/push"
+        self.harness.charm._loki_consumer.loki_endpoints = []
```
mmanciop, line 259: I don't think we should use mocks for these tests, it makes me very uncomfortable

mmanciop (changes requested):
Where are the tests?

tests/unit/test_charm.py
mmanciop: For evidence, these mocks have hidden this bug all along https://github.com/canonical/loki-k8s-operator/pull/138

tests/unit/test_charm.py
Abuelodelanada: We will address the removal of all these mocks in https://github.com/canonical/grafana-agent-k8s-operator/issues/45

tests/unit/test_charm.py
Abuelodelanada: We will address the removal of all these mocks in https://github.com/canonical/grafana-agent-k8s-operator/issues/45

dstathis (approved):

tests/unit/test_charm.py
```diff
@@ -267,6 +267,52 @@ def test__on_loki_push_api_endpoint_departed(
         self.assertEqual(path, "/etc/agent/agent.yaml")
         self.assertTrue(yaml.safe_load(content)["loki"] == {})
 
+    def test__loki_config_no_endpoints(self):
+        consumer = self.harness.charm._loki_consumer
+        endpoints = PropertyMock(return_value="")
+        type(consumer).loki_endpoints = endpoints
```
sed-i, line 273: Would `patch.object` work here instead of direct monkeypatching?

sed-i (approved):

tests/unit/test_charm.py
mmanciop: IMO there is zero reason to mock here at all. We have had this discussion after the daily yesterday. This situation can be modelled using the relation-related capabilities of `Harness`, and a test written with `Harness` is way more valuable than one using mocks, because mocks are brittle, while the relations test the actual charm's contract :-)

simskij:
Tests currently failing. Waiting to do a final review until it passes again.

tests/unit/test_charm.py
```diff
@@ -85,6 +83,7 @@ def setUp(self):
 
     @responses.activate
     def test_remote_write_configuration(self):
+        self.harness.set_can_connect("agent", True)
```
sed-i, line 86: The `setUp` calls `begin_with_initial_hooks`, so it seems natural to me to add `self.harness.container_pebble_ready("agent")` immediately after, instead of `set_can_connect(True)`.
A test that specifically tests for False could do `set_can_connect(False)` in that one place even when we have `container_pebble_ready("agent")` at the top.

tests/unit/test_charm.py
```diff
@@ -269,3 +214,47 @@ def test__on_loki_push_api_endpoint_departed(self):
...
+        endpoint1 = json.dumps({"url": "http://loki1:3100:/loki/api/v1/push"})
+        self.harness.update_relation_data(rel_id, "loki/1", {"endpoint": endpoint1})
+
+        expected = {
+            "loki": {
+                "configs": [
+                    {
+                        "name": "promtail",
+                        "clients": [
+                            {"url": "http://loki0:3100:/loki/api/v1/push"},
```
sed-i, line 236: `loki0:3100:` seems invalid, which makes me wonder if url validation should take place in the charm/lib.

tests/unit/test_charm.py
```diff
@@ -269,3 +214,47 @@ def test__on_loki_push_api_endpoint_departed(self):
...
+        self.harness.set_can_connect("agent", True)
+        rel_id = self.harness.add_relation("logging-consumer", "loki")
+
+        self.harness.add_relation_unit(rel_id, "loki/0")
+        endpoint0 = json.dumps({"url": "http://loki0:3100:/loki/api/v1/push"})
+        self.harness.update_relation_data(rel_id, "loki/0", {"endpoint": endpoint0})
+
+        self.harness.add_relation_unit(rel_id, "loki/1")
+        endpoint1 = json.dumps({"url": "http://loki1:3100:/loki/api/v1/push"})
+        self.harness.update_relation_data(rel_id, "loki/1", {"endpoint": endpoint1})
```
sed-i, line 228:
```suggestion
        for u in range(2):
            self.harness.add_relation_unit(rel_id, f"loki/{u}")
            endpoint = json.dumps({"url": f"http://loki{u}:3100/loki/api/v1/push"})
            self.harness.update_relation_data(rel_id, f"loki/{u}", {"endpoint": endpoint})
```

tests/unit/test_charm.py
Abuelodelanada: good catch!

tests/unit/test_charm.py
Abuelodelanada: In this test we are checking that the URLs sent by the Provider side of the relation are obtained by the Consumer side...

I think there is no need to validate URL since we are using `socket.getfqdn()`:

```python
    @property
    def _url(self) -> str:
        """Get local Loki Push API url.

        Return url to loki, including port number, but without the endpoint subpath.
        """
        return "http://{}:{}".format(socket.getfqdn(), self.port)
```

dstathis (approved):

src/charm.py
```diff
@@ -193,6 +193,8 @@ def _update_config(self, event=None):
                 self.unit.status = ActiveStatus()
         except GrafanaAgentReloadError as e:
             self.unit.status = BlockedStatus(str(e))
+        except APIError as e:
+            self.unit.status = WaitingStatus(str(e))
```
sed-i, line 197:
```suggestion
        except APIError as e:
            # The "agent" service was not created yet. Need to wait for pebble-ready.
            self.unit.status = WaitingStatus(str(e))
```

sed-i (approved):

PR #50, opened 2022-06-24
promote charm action

dstathis:
Github action to promote a release to a different channel. Testing it out here before copying over to other repos.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@
 [tox]
 skipsdist=True
 skip_missing_interpreters = True
-envlist = lint, static-{charm,lib,unit,integration}, unit
+envlist = lint, static-{charm,lib}, unit
 
 [vars]
 src_path = {toxinidir}/src
--- a/.github/workflows/promote.yaml
+++ b/.github/workflows/promote.yaml
@@ -1,0 +1,31 @@
+name: Promote Charm
+
+on:
+  workflow_dispatch:
+    inputs:
+      promote-to:
+        type: choice
+        description: Channel to promote to
+        options:
+          - beta
+          - stable
+      promote-from:
+        type: choice
+        description: Channel to promote from
+        options:
+          - beta
+          - edge
+
+jobs:
+  promote:
+    name: Promote Charm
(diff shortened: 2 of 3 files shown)


rbarry82:
`mypy` strikes again.

We actually got the go-ahead (in Prometheus, and the patch I'm working on for loki alerting here) to drop static analysis of `tests/`. Just remove the tox target for `static-[unit|integration]`

rbarry82 (dismissed):

tox.ini
```diff
@@ -54,11 +54,8 @@ commands =
     isort --check-only --diff {[vars]all_path}
     black --check --diff {[vars]all_path}
 
-[testenv:static-{charm,lib,unit,integration}]
```
sed-i, line 57: I think it's nice to have the option to run them when we want to.
Perhaps remove them from the [workflow](https://github.com/canonical/grafana-agent-k8s-operator/blob/main/.github/workflows/pull-request.yaml) instead of from tox?

dstathis: Once there is one error, though, it will just fail forever right?

sed-i: Yes, it will fail until someone decides to do something about it :)
I plan to keep using it, unless there is a compelling reason to remove it from tox in addition to the workflow.

sed-i (approved):

rbarry82 (approved):

tox.ini
dstathis: While there may not be a workaround. I think we decided to leave it in tox.

sed-i: Currently, `git stash pop` for 
```difff
-    mypy
+    mypy==0.950
+    httpcore==0.14.7
+    responses==0.20.0
```
is quick and dirty and works.

If we deem the tox env as necessary/useful I am happy to pin mypy for now, just like we do for many other packages.
```
find .. -type f -name "requirements*.txt" -or -name "tox.ini" -print0 | xargs -0 rg "==|>|<"
```
I think mypy 0.95 is better than no mypy. At least for now.

PR #51, opened 2022-06-27
Fix promote action.

dstathis:
Typo and change the default channel.

### DIFF
--- a/.github/workflows/promote.yaml
+++ b/.github/workflows/promote.yaml
@@ -34,5 +34,5 @@ jobs:
         with:
           credentials: ${{ secrets.CHARMCRAFT_CREDENTIALS }}
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          destination-channel: ${{ env.promote-to }}
-          origin-channel: ${{ github.event.inputs.promote-from }}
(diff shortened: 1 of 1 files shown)


sed-i (dismissed):
This kind of thing is "tested in production", right?

simskij (dismissed):

.github/workflows/promote.yaml
```diff
@@ -8,13 +8,15 @@ on:
...
+          - candidate
           - stable
       promote-from:
         type: choice
         description: Channel to promote from
         options:
-          - beta
           - edge
+          - beta
+          - candidate
```
sed-i, line 19: Seems like it would be possible to "promote" from stable to candidate.
Is that intended?
Should we instead have only one input - "Channel to promote"?

dstathis: I'm not sure I understand. Stable is not a choice for "promote-from". I don't see how we could have only one.

dstathis: Okay I see what you're saying here. I think there could be times where we promote up two levels. I think the flexibility would be nice. What do others think?

simskij (changes requested):

simskij (approved):

PR #54, opened 2022-07-13
Forward Loki alert rules as well as Prometheus alert rules

rbarry82:
## Issue
For completeness, Grafana Agent should forward both types of alerts to the appropriate endpoints.

In addition, testing this required building out the unit tests to a point where they are meaningful. We were already forwarding Prometheus rules, but no tests were present.

## Solution
Create partial methods on `GrafanaAgentOperatorCharm` which serve as observers with some default args, which then carry through to an `update_alerts` method which dynamically handles dispatching/reloading. Use a namedtuple for `src` and `dest` for a little abstraction.

## Testing Instructions
Relate and verify

## Release Notes
Forward Loki alert rules as well


Abuelodelanada (approved):

src/charm.py
```diff
@@ -99,18 +118,42 @@ def __init__(self, *args):
...
+    def _loki_alerts(self, event):
+        self.update_alerts_rules(
+            event,
+            alerts_func=self._loki_provider.alerts,
+            reload_func=self._loki_consumer._reinitialize_alert_rules,
+            mapping=self.loki_rules_paths,
+        )
+
+    def update_alerts_rules(
+        self, _, alerts_func: Any, reload_func: Callable, mapping: RulesMapping
```
balbirthomas, line 138: Bit surprising to see a positional argument that is not being used. I presume it is because this function definition needs to confirm to some function signature.

src/charm.py
```diff
@@ -99,18 +118,42 @@ def __init__(self, *args):
...
+    ):
         """Copy alert rules from relations and save them to disk."""
-        rules = self._scrape.alerts()
-        shutil.rmtree(self._metrics_rules_dest_path)
-        shutil.copytree(self._metrics_rules_src_path, self._metrics_rules_dest_path)
+        rules = {}
+
+        # MetricsEndpointConsumer.alerts is not @property, but Loki is, so
+        # do the right thing
+        if callable(alerts_func):
```
balbirthomas, line 145: Interesting I did not realize we had this inconsistency between our charms. Not relevant to this PR it leaves me wondering if it worth making `alerts` in Loki and Prometheus consistent, which ever way.

tests/integration/loki-tester/requirements.txt
```diff
@@ -0,0 +1,2 @@
+ops >= 1.2.0
```
balbirthomas, line 1: There is an integration test failure. I think `tox.ini` may need 'aiohttp'.

balbirthomas (approved):

sed-i (commented):
Do we expect to be forking the two tester charms more than just here?
If yes, perhaps turn them into relation interface (tester) charms?

src/charm.py
rbarry82: It's the actual event, which we ignore in a surprising amount of charms, it's just really obvious here.

rbarry82:
> Do we expect to be forking the two tester charms more than just here? If yes, perhaps turn them into relation interface (tester) charms?

I wouldn't really call this a fork either, just a copy+paste because they already did what we needed. I definitely don't think we should turn them in any kind of "real" charms (relation interface tester or other). Both tester charms, both here and in their "original" repos, are cleaned, new client libraries copied in, and rebuilt when the integration tests are run.

They are essentially dynamically generated, and they should stay there.

PR #55, opened 2022-08-15
update image

dstathis:
## Issue
Update the image to the latest version

## Solution
update in metadata.yaml and test

## Context
Ran basic manual testing and integration tests

## Testing Instructions
deploy and fiddle

## Release Notes
updated grafana-agent to version 0.26.1

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -23,7 +23,7 @@ containers:
 resources:
   agent-image:
     type: oci-image
-    upstream-source: grafana/agent:v0.20.1
+    upstream-source: grafana/agent:v0.26.1
     description: OCI image for Grafana Agent
 
 requires:
--- a/src/charm.py
+++ b/src/charm.py
@@ -259,7 +259,7 @@ class GrafanaAgentOperatorCharm(CharmBase):
         Returns:
             The arguments as a string
         """
-        return "-config.file=/etc/agent/agent.yaml -prometheus.wal-directory=/tmp/agent/data"
+        return "-config.file=/etc/agent/agent.yaml"
 
     def _config_file(self) -> Dict[str, Any]:
         """Generates config file str.
@@ -351,7 +351,8 @@ class GrafanaAgentOperatorCharm(CharmBase):
             The dict representing the config
         """
         return {
-            "prometheus": {
+            "metrics": {
+                "wal_directory": "/tmp/agent/data",
                 "configs": [
                     {
                         "name": "agent_scraper",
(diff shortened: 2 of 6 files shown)


rbarry82:
LGTM other than the failing linters, assuming integration tests pass

metadata.yaml
```diff
@@ -23,7 +23,7 @@ containers:
 resources:
   agent-image:
     type: oci-image
-    upstream-source: grafana/agent:v0.20.1
+    upstream-source: grafana/agent:v0.26.1
```
sed-i, line 26: Update README as well?
What do you think about a resource revision table ([example](https://github.com/canonical/alertmanager-k8s-operator/blob/main/README.md#resource-revisions))?

src/charm.py
```diff
@@ -351,7 +351,8 @@ def _prometheus_config(self) -> dict:
             The dict representing the config
         """
         return {
-            "prometheus": {
+            "metrics": {
```
sed-i, line 354: The [official example](https://grafana.com/docs/agent/latest/configuration/create-config-file/) has `remote_write` under `global` as well as under `configs`.
Any idea which one we need? Makes more sense to me that the `remote_write` section needs to be under `global`.

sed-i: Similarly, for the loki portion of the config, [looks like](https://grafana.com/docs/agent/latest/configuration/create-config-file/) we now need `logs` instead of `loki` for the section name.
Docs [seem partial](https://grafana.com/docs/agent/latest/configuration/create-config-file/#full-config-example) but would be nice to confirm the new format for `logs.scrape_configs` section supports what we currently have coded.

metadata.yaml
rbarry82: We should actually remove that from Alertmanager instead. Having a resource revision table is practically begging users to deploy mismatched charm+workload.

src/charm.py
rbarry82: `remote_write` under `global` sends _all_ metrics there. `metrics` is explicitly:

> The Prometheus subsystem config is useful for those migrating from Prometheus and _those who want to scrape metrics from something that currently does not have an associated integration._

^^ this is our use case (at least partly).

`loki` [is in fact deprecated](https://github.com/grafana/agent/blob/main/pkg/config/config.go#L102) the same as `metrics` (and `tempo`, which we don't use), but still works for now. There's probably a `DEPRECATION NOTICE:` in the container logs. We may as well change it, but it's ok for now

rbarry82 (approved):

metadata.yaml
dstathis: I agree, it's asking for trouble.

src/charm.py
dstathis: I changed it.

PR #58, opened 2022-08-18
Updating documentation

Abuelodelanada:
## Issue

This PR upgrades this charm's documentation.


INTEGRATING.md
```diff
@@ -1,36 +1,171 @@
...
+
+The information exchanged through the `loki_push_api` interface can be broken down into two parts:
+
+  - Grafana Agent charm provides an [endpoint URL](https://grafana.com/docs/loki/latest/api/#post-lokiapiv1push) to receive log from [Loki clients](https://grafana.com/docs/loki/latest/clients/) that relates with this charm.
+  - Grafana Agent charm may also receive alert rules which tell when to raise alerts. These rules are read from a directory named `loki_alert_rules`, if present at the top level, within the client charm's source (`src`) directory.
 
-  - Loki may receive alert rules which tell when to raise alerts. These rules are read from a directory named `loki_alert_rules`, if present at the top level, within the client charm's source (`src`) directory.
 
+For instance let's say that we have a [Zinc Charmed Operator](https://charmhub.io/zinc-k8s) that implements the other side (`requires`) of the relation.
+After deploying this charm, we can relate `Grafana Agent` and `Zinc` through `loki_push_api` relation interface:
```
sed-i, line 26: Need to clearly differentiate between the charm's function as provider and as requirer of the same relation, but how about:

> To have a charm send logs to loki, you can use one of the "consumer" object offered by the `loki_push_api`.
> 
> First, specify the relation in the charm's metadata.yaml:
> 
> ```yaml
> requires:
>   logging:
>     interface: loki_push_api
> ```
> 
> Next, obtain the library from charmhub:
> 
> ```shell
> charmcraft fetch-lib charms.loki_k8s.v0.loki_push_api
> ```
> 
> Then, use the charm library in the charm:
> 
> ```python
> self._logging = LogProxyConsumer(...)
> # or
> self._logging = LokiPushApiConsumer(...)
> ```
> 
> Finally, relate the app (e.g. zinc) to loki:

INTEGRATING.md
```diff
@@ -1,36 +1,171 @@
...
+```
+
+### Self metrics endpoint
+
+
+```yaml
+self-metrics-endpoint:
+    interface: prometheus_scrape
+```
+This Grafana Agent charm may forward information about its metrics endpoint and associated alert rules to a Prometheus charm over the `self-metrics-endpoint` relation using the [`prometheus_scrape`](https://charmhub.io/prometheus-k8s/libraries/prometheus_scrape) interface. In order for these metrics to be aggregated by the remote Prometheus charm all that is required is to relate the two charms as in:
```
sed-i, line 59:
```suggestion
Grafana agent has its own metrics endpoint and the charm comes with associated alert rules, which can be used for self-monitoring. For this purpose the charm has the `self-metrics-endpoint` relation using the [`prometheus_scrape`](https://charmhub.io/prometheus-k8s/libraries/prometheus_scrape) interface. In order for these metrics to be aggregated by the remote Prometheus charm all that is required is to relate the two charms as in:
```

INTEGRATING.md
```diff
@@ -1,36 +1,171 @@
...
+```
+
+
+
+### Metrics endpoint
+
+```yaml
+  metrics-endpoint:
+    interface: prometheus_scrape
+```
```
sed-i, line 91: I wonder if we should change the format a bit so things are more obvious and more copy-pasteable.
For example:

> ### Metrics endpoint
> Charms may forward information about their metrics endpoints and associated alert rules to the Grafana Agent charm over the `metrics-endpoint` relation using the [`prometheus_scrape`](https://charmhub.io/prometheus-k8s/libraries/prometheus_scrape) interface.
> 
> #### This charm
> ```yaml
> requires:
>   metrics-endpoint:
>     interface: prometheus_scrape
> ```
> 
> #### Remote charm
> 
> ```yaml
> provides:
>   metrics-endpoint:
>     interface: prometheus_scrape
> ```

etc.

INTEGRATING.md
```diff
@@ -1,36 +1,171 @@
...
 ```mermaid
 flowchart LR
     subgraph COS Lite
         prometheus
         loki
+        grafana
     end
     subgraph K8s cluster
         prometheus               ---      |remote_write| grafana-agent
+        prometheus               ---      |prometheus_scrape| grafana-agent
```
sed-i, line 166: Iiuc, this line shouldn't be here:
- Both prom and agent have `metrics-endpoint` under requires
- The agent is "converting" scrape targets into remote write

INTEGRATING.md
```diff
@@ -1,36 +1,171 @@
...
+### Logging provider
+
+```yaml
+  logging-provider:
+    interface: loki_push_api
+```
+
+Grafana Agent Charmed Operator may receive logs from any charm that supports the [`loki_push_api`](https://charmhub.io/loki-k8s/libraries/loki_push_api) relation interface.
+
+Besides, the `logging-provider` relation can inject and setup `Promtail` into the workload container of the consumer side of the relation.
```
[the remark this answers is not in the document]
Abuelodelanada: Good catch!. This paragraph is wrong. Thanks!

INTEGRATING.md
Abuelodelanada: I like it!

INTEGRATING.md
Abuelodelanada: I like this format!

INTEGRATING.md
Abuelodelanada: In fact this `prmetheus_scrape` is the interface used by `self-metrics-endpoint` relation, with this Prometheus can scrape Grafana Agent self metrics.

Should we show relation names too?

Abuelodelanada: Maybe we should not show self-metrics relations...

sed-i: Yeah I think in the context of this deployment scenarios it is unnecessary.
I would certainly add the self monitoring relations in the bundle doc though.

Other kinds of self monitoring though could be worth mentioning here, but I think it would be a separate diagram. For example the resilience @rbarry82 was mentioning a while ago.

INTEGRATING.md
```diff
@@ -1,36 +1,292 @@
...
+    interface: loki_push_api
+```
+
+2. Obtain the library from charmhub:
+
+```shell
+charmcraft fetch-lib charms.loki_k8s.v0.loki_push_api
+```
+
+3. Import the library and use it in `src/cham.py`:
```
sed-i, line 81:
```suggestion
3. Import the library and use it in your `src/charm.py`:
```

sed-i (approved):
Nice!

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+  logging:
+    interface: loki_push_api
+```
+
+Using this relation, Grafana Agent Charmed Operator can receive logs from any charm that supports the [`loki_push_api`](https://charmhub.io/loki-k8s/libraries/loki_push_api) relation interface (Those logs will later be forwarded to Loki using the `logging-consumer` relation).
+
+
+The information exchanged through the `loki_push_api` interface can be broken down into two parts:
+
+  - Grafana Agent charm provides an [endpoint URL](https://grafana.com/docs/loki/latest/api/#post-lokiapiv1push) to receive log from charms running [Loki clients](https://grafana.com/docs/loki/latest/clients/).
```
dstathis, line 60:
```suggestion
  - Grafana Agent charm provides an [endpoint URL](https://grafana.com/docs/loki/latest/api/#post-lokiapiv1push) to receive logs from charms running [Loki clients](https://grafana.com/docs/loki/latest/clients/).
```

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+  - Grafana Agent charm may also receive alert rules which tell when to raise alerts. These rules are read from a directory named `loki_alert_rules`, if present at the top level, within the client charm's source (`src`) directory.
```
dstathis, line 61:
```suggestion
  - Grafana Agent charm may also receive alert rules which tell it when to raise alerts. These rules are read from a directory named `loki_alert_rules`, if present at the top level, within the client charm's source (`src`) directory.
```

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+
+
+For instance let's say that we need [Zinc Charmed Operator](https://charmhub.io/zinc-k8s) send logs to Grafana Agent.
```
dstathis, line 64:
```suggestion
For instance let's say that we need [Zinc Charmed Operator](https://charmhub.io/zinc-k8s) to send logs to Grafana Agent.
```

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+To do that we can use one of the "consumer" object offered by the `loki_push_api` lib and follow these steps.
```
dstathis, line 65:
```suggestion
To do that we can use one of the "consumer" objects offered by the `loki_push_api` lib and follow these steps.
```

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+
+#### Remote charm
+
+```yaml
+requires:
+  grafana-dashboard:
+    interface: grafana_dashboard
+```
+
+Over the `grafana-dashboard` relation using the [`grafana-dashboard`](https://charmhub.io/grafana-k8s/libraries/grafana_dashboard) interface, this Grafana Agent charm also provides meaningful dashboards (placed in `src/grafana_dashboards`) about its metrics to be shown in a [Grafana Charm ](https://charmhub.io/grafana-k8s). These dashboards are placed
```
dstathis, line 117:
```suggestion
Over the `grafana-dashboard` relation using the [`grafana-dashboard`](https://charmhub.io/grafana-k8s/libraries/grafana_dashboard) interface, this Grafana Agent charm also provides meaningful dashboards (placed in `src/grafana_dashboards`) about its metrics to be shown in a [Grafana Charm ](https://charmhub.io/grafana-k8s).
```

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+```
+
+Charms may forward information about their metrics endpoints and associated alert rules to the Grafana Agent charm over the `metrics-endpoint` relation using the [`prometheus_scrape`](https://charmhub.io/prometheus-k8s/libraries/prometheus_scrape) interface.
+
+```shell
+juju relate grafana-agent-k8s:metrics-endpoint zinc-k8s
+```
+
+Charms that seek to provide metrics endpoints and alert rules must do so using the provided [`prometheus_scrape`](https://charmhub.io/prometheus-k8s/libraries/prometheus_scrape) charm library.  This library by implementing the `metrics-endpoint` relation, not only ensures that scrape jobs and alert rules are forward to Grafana Agent but also that these are updated any time the metrics provider charm is upgraded. For example new alert rules may be added or old ones removed by updating and releasing a new version of the metrics provider charm. While it is safe to update alert rules as desired, care must be taken when updating scrape job specifications as this has the potential to break the continuity of the scraped metrics time series. In particular changing the following keys in the scrape job can break time series continuity
```
dstathis, line 207:
```suggestion
Charms that seek to provide metrics endpoints and alert rules must do so by using the provided [`prometheus_scrape`](https://charmhub.io/prometheus-k8s/libraries/prometheus_scrape) charm library.  This library, by implementing the `metrics-endpoint` relation, not only ensures that scrape jobs and alert rules are forwarded to Grafana Agent but also that these are updated any time the metrics provider charm is upgraded. For example new alert rules may be added or old ones removed by updating and releasing a new version of the metrics provider charm. While it is safe to update alert rules as desired, care must be taken when updating scrape job specifications as this has the potential to break the continuity of the scraped metrics time series. In particular changing the following keys in the scrape job can break time series continuity
```

INTEGRATING.md
```diff
@@ -1,36 +1,291 @@
...
+
+Grafana Agent may forward client charms metrics and associated alert rules (that are received using the `metrics-endpoint` relation) to Prometheus using the [`prometheus_remote_write`](https://charmhub.io/prometheus-k8s/libraries/prometheus_remote_write) interface.
+To do that you have to relate Grafana Agent with Prometheus:
+
+```bash
+juju relate grafana-agent-k8s:send-remote-write prometheus-k8s
+```
+
+
+Once these relations are established, you can check that by running:
```
dstathis, line 242:
```suggestion
Once these relations are established, you can check it by running:
```

README.md
```diff
@@ -1,17 +1,23 @@
...
-[Grafana Agent](https://github.com/grafana/agent) is a telemetry collector for
-sending metrics, logs, and trace data to the opinionated Grafana observability
-stack.
 
-As a single entry point to the observability stack, the grafana agent charm
+[Grafana Agent](https://github.com/grafana/agent) is a telemetry collector for sending metrics,
+logs, and trace data to the opinionated Grafana observability stack.
+
+The Grafana Agent Charmed Operator deploys Grafana Agent in Kubernetes using [Juju](https://juju.is)
+the [Charmed Operator Lifecycle Manager (OLM)](https://juju.is/docs/olm).
```
dstathis, line 12:
```suggestion
and the [Charmed Operator Lifecycle Manager (OLM)](https://juju.is/docs/olm).
```

tests/unit/test_alerts.py
```diff
@@ -160,7 +160,7 @@ def test_forwards_prometheus_rules(self):
         rules = json.loads(data["alert_rules"])
 
         for group in rules["groups"]:
-            if group["name"].endswith("provider-tester_alerts"):
+            if group["name"].endswith("provider-tester_alerts_alerts"):
```
dstathis, line 163: Is this a bug? I think the group should not end with "_alerts_alerts".

Abuelodelanada: mmm maybe it's a breaking change in the lib... I'll check it

Abuelodelanada: Hi @rbarry82 !

I found a bug in the regex changed [here](https://github.com/canonical/prometheus-k8s-operator/commit/d5d034182188f95b55990a6a6195c0656cdb3f50#diff-a15ebc3aa16d0affd9ee815751bc53054676385cd1f73bfc36bc3ad10b1c0da3R285)

With this change, the name of the alert group change from:

`lma_f2c1b2a6-e006-11eb-ba80-0242ac130004_provider-tester_alerts`

to:

`lma_12345678_grafana-agent-k8s_lma_f2c1b2a6-e006-11eb-ba80-0242ac130004_provider-tester_alerts_alerts`

because of [this piece of code](https://github.com/canonical/prometheus-k8s-operator/blob/main/lib/charms/prometheus_k8s/v0/prometheus_remote_write.py#L232):

```python
                if not self._is_already_modified(alert_group["name"]):
                    # update group name with topology and sub-path
                    alert_group["name"] = self._group_name(
                        str(root_path),
                        str(file_path),
                        alert_group["name"],
                    )
```

It is duplicating the final `_alerts` in the name.

should we revert the regex or the use case changed?

PR #59, opened 2022-08-31
Upkeep

dstathis:
## Issue
General charm upkeep.

## Solution
Consolidate CI and set workload version.
The way I did the unit test mocking seems a bit ugly to me. Is there a cleaner way?

## Testing Instructions
Deploy and check workload version is set.

## Release Notes
grafana-agent-k8s now sets workload version


src/charm.py
```diff
@@ -200,6 +201,10 @@ def on_pebble_ready(self, _: EventBase) -> None:
         self._container.add_layer(self._name, pebble_layer, combine=True)
         self._container.autostart()
 
+        if (version := self._agent_version) is not None:
+            self.unit.set_workload_version(version)
+        else:
+            logger.debug("Could not get agent version")
```
sed-i, line 207: This may look suspicious when looking at the debug log. Clarify it's "not a big deal"?

sed-i (approved):
Can you please check if you could use anything from #56 and close it?

src/charm.py
dstathis: Done.

src/charm.py
```diff
@@ -416,6 +421,18 @@ def _reload_config(self, attempts: int = 10) -> None:
...
+    def _agent_version(self) -> str:
+        """Returns the version of the agent.
+
+        Returns:
+            A string equal to the agent version
+        """
+        if not self._container.can_connect():
+            return None
+        version_output, _ = self._container.exec(["/bin/agent", "-version"]).wait_output()
+        return re.search(r"v(\d*\.\d*\.\d*)", version_output).group(1)
```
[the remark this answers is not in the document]
dstathis: Done.

src/charm.py
```diff
@@ -431,7 +433,12 @@ def _agent_version(self) -> Optional[str]:
         if not self._container.can_connect():
             return None
         version_output, _ = self._container.exec(["/bin/agent", "-version"]).wait_output()
-        return re.search(r"v(\d*\.\d*\.\d*)", version_output).group(1)  # type: ignore
+        # Output looks like this:
+        # agent, version v0.26.1 (branch: HEAD, revision: 2b88be37)
+        result = re.search(r"v(\d*\.\d*\.\d*)", version_output)
```
rbarry82, line 438: Is there a reason why we're stripping off the `v` at the beginning? It probably doesn't make a semantic difference, but we're a little hemmed in here.

Not that I ever really expect this to fail, but we're also checking to see if the container exec fails.

Either way, `v..` actually passes this regexp. I always personally prefer `groupdict` for legibility, even if it adds verbosity. Raw strings also great for regexps.

`re.search(r`version\s+(?P<version>.*?)\s+`).groupdict().get("version", None)` should do the same thing as a oneliner which is more or less immune to changes and grabs whatever whitespace-delimited string comes after "version", or `None` if nothing was matched.

dstathis: I find it a bit more simple as is so I will leave it for now. If the regex ever gets more complicated, we can change it as necessary.

PR #62, opened 2022-10-18
Remove self-metrics-endpoint and grafana-dashboard relations

Abuelodelanada:
## Issue

The PR removes `self-metrics-endpoint` and `grafana-dashboard`

Grafana Agent is not scrapable, seems we added this by mistake


sed-i (approved):
Would be nice to include the reason for the removal in the PR description. It's not immediately obvious to me.

simskij (approved):

rbarry82 (approved):

PR #66, opened 2022-11-16
Create grafana_agent library

dstathis:
This PR pulls out common code which can be used for both k8s and machine charms and creates a `grafana_agent` charm library.


simskij:
Before getting to addressing the content of the PR (which looks fine from a glance), it struck me: do we really need a library for this or should we just split it into a base class, in its own file, in the `src` folder? We could use this repo for both flavors, couldn't we?

dstathis:
> Before getting to addressing the content of the PR (which looks fine from a glance), it struck me: do we really need a library for this or should we just split it into a base class, in its own file, in the `src` folder? We could use this repo for both flavors, couldn't we?

I'm not exactly sure how that would work. We would need two different `charmcraft.yaml` files and two different `src/charm.py` files. I don't think charmcraft can really handle that.

tox.ini
```diff
@@ -34,6 +34,20 @@ commands =
     isort {[vars]all_path}
     black {[vars]all_path}
 
+[testenv:render-k8s]
+description = Render the k8s charm
+allowlist_externals = cp
+commands =
+    cp src/k8s_charm.py src/charm.py
+    cp k8s_metadata.yaml metadata.yaml
```
sed-i, line 42: By default, charmcraft includes everything under `src/`. May want to exclude. Not critical.

Iirc, this won't readily work from other dirs, unless
```suggestion
    cp {toxinidir}/src/k8s_charm.py {toxinidir}/src/charm.py
    cp {toxinidir}/k8s_metadata.yaml {toxinidir}/metadata.yaml
```

simskij:
> > Before getting to addressing the content of the PR (which looks fine from a glance), it struck me: do we really need a library for this or should we just split it into a base class, in its own file, in the `src` folder? We could use this repo for both flavors, couldn't we?
> 
> I'm not exactly sure how that would work. We would need two different `charmcraft.yaml` files and two different `src/charm.py` files. I don't think charmcraft can really handle that.

We'll solve that using `tox` environments, something along the lines of:

```bash
$ tox -e build-machine

# or

$ tox -e build-k8s
```

and copy different `metadata.yaml` files in depending on the env.

src/grafana_agent.py
```diff
@@ -1,16 +1,33 @@
...
+share and consume charm libraries. They serve to enhance collaboration
+between charmers. Use a charmer's libraries for classes that handle
+integration with their charm.
+
+Bear in mind that new revisions of the different major API versions (v0, v1,
+v2 etc) are maintained independently.  You can continue to update v0 and v1
+after you have pushed v3.
+
+Markdown is supported, following the CommonMark specification.
+"""
```
sed-i, line 23:
```suggestion
"""Common logic for both k8s and machine charms for Grafana Agent."""
```

src/grafana_agent.py
```diff
@@ -20,14 +37,22 @@
...
+# The unique Charmhub library identifier, never change it
+LIBID = "797d2fe4842f44c99106923e037f237f"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
```
sed-i, line 55: No longer a lib, right?
```suggestion
```

src/grafana_agent.py
```diff
@@ -119,6 +113,39 @@ def __init__(self, *args):
             self._on_loki_push_api_endpoint_departed,
         )
 
+    # Abstract Methods
+
+    def agent_version_output(self) -> str:
```
PietroPasotti, line 118: Hate to say this now, but now that I see how the code looks like with this pattern, I kind of think I'd like it more if we used a different pattern. 
Instead of abstract charm + inheritance, I'd go for concrete charm + composition + two separate grafana controller classes.

Why:
- charm code is complicated enough without throwing inheritance into the mix. Suddenly we have three charms to think about.
- I liked the pattern in loki https://github.com/canonical/loki-k8s-operator/blob/main/src/loki_server.py where we have a separate LokiServer class. 

So I'd rather have a single charm class (no subclasses) handling the operation logic, and sending out calls to a GrafanaServer (or whatever) class, that is either a GrafanaServerK8s or a GrafanaServerMachine.
So the charm code becomes simpler:
```
if self.grafana_server.is_ready(): ...
```

We can also say: too late, that ship has sailed, and I'll shut up and review this PR instead.

dstathis: Both solutions seem to be more or less the same to me and I'd rather not rewrite it. We can discuss further with the whole team.

src/grafana_agent.py
```diff
@@ -52,32 +49,50 @@ def __init__(self, message="could not reload configuration"):
         super().__init__(self.message)
 
 
-class GrafanaAgentOperatorCharm(CharmBase):
+class GrafanaAgentBase(Object):
     """Grafana Agent Charm."""
```
PietroPasotti, line 53: Not a charm
```suggestion
    """Grafana Agent object."""
```

src/grafana_agent.py
```diff
@@ -210,58 +195,58 @@ def on_remote_write_changed(self, event: RelationChangedEvent) -> None:
...
-        self.unit.status = ActiveStatus()
+        self.charm.unit.status = ActiveStatus()
 
     def _update_config(self, _) -> None:
-        if not self._container.can_connect():
-            # Pebble is not ready yet so no need to update config
-            self.unit.status = WaitingStatus("waiting for agent container to start")
+        if not self._is_ready():
+            # Grafana-agent is not yet running so no need to update config
+            self.charm.unit.status = WaitingStatus("waiting for agent to start")
```
PietroPasotti, line 212: so now we have an object other than the charm manipulating the charm's status? Seems like a risky pattern to me.

PietroPasotti (changes requested):
I only gave a very quick look at the new pattern, and I'm not sure we're there yet.

We had two charms inheriting from a base charm class.
Your new, current proposal is to have two charms (or just one), using one of two objects to manage their own state and the grafana agent process.

To me that feels like violating the principle of separation of concerns.
When I proposed composing the grafana agent class instead of inheriting, what I had in mind was  something like:

```
class GrafanaAgent:
    """Substrate-agnostic grafana agent code.""""
    def restart_if_healthy(self) -> bool:
        if self.healthy():
            self.restart()
            return True
        return False

    def restart(): raise NotImplementedError('substrate-specific')

    def healthy(): raise NotImplementedError('substrate-specific')    
    
class GrafanaAgentK8s(GrafanaAgent):
    """K8s-specific grafana agent code.""""
    @property
    def container() -> Container: ... 
    
    def healthy(self) -> bool:
        self.container.exec(...)
        
    def restart(self) -> bool:
        self.container.replan()
        
class GrafanaAgentMachine(GrafanaAgent):
    """Machine-specific grafana agent code.""""
     def healthy(self) -> bool:
        Popen([...])
        
    def restart(self) -> bool:
        Popen([...])
        
class GrafanaAgentCharm(CharmBase):
    """Grafana Agent Charm"""
    def __init__(...):
        self.agent = GrafanaAgentMachine() if get_substrate() == 'machine' else GrafanaAgentK8s()
        
     def _on_whatever(self, event):
         success = self.agent.restart_if_healthy()
         if success: 
             self.status = Active...
         else:
             self.status = Blocked...
```

Not sure if this makes things more clear at all, but my hope/idea/intuition was that we'd have GrafanaAgent classes (one for each supported substrate) that know nothing about the charm -- they just know how to perform (substrate-specific) process (micro)management tasks (what it means to 'restart grafana agent', or 'purge the config', or 'kill it', etc...), and report the status of the process.
Based on the outcome of those actions and the process status, the charm does all meta-level ops stuff.

That way we can keep the ops isolated at the charm level, and the process management at the lower level. 

Exactly like https://github.com/canonical/grafana-k8s-operator/blob/main/src/grafana_server.py does it. GrafanaServer knows nothing about charms, juju, statuses, etc... It just knows how to grab stuff from a running process inside a container. In our case it's the same, only we have a separate class for the case in which the process is not in a container, but running on 'this machine'.

src/grafana_agent.py
PietroPasotti: see leading comment for the context

src/k8s_charm.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        self._container.push(path, text)
+
+    def restart(self) -> None:
+        """Restart grafana agent."""
+        self._container.restart("agent")
+
+
+if __name__ == "__main__":
+    main(GrafanaAgentK8sCharm)
```
sed-i, line 94: I probably missed a few discussions about this but how about:

```python
def is_this_k8s() -> bool:
    import os
    return bool(os.environ.get("KUBERNETES_SERVICE_HOST"))
    # or JUJU_MACHINE_ID
    # or KUBERNETES_PORT

if __name__ == "__main__":
    main(GrafanaAgentK8sCharm if is_this_k8s() else GrafanaMachineCharm)
```

metadata.yaml
```diff
@@ -1,6 +1,8 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 
+# Stub file to make CI work correctly
```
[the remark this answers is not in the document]
dstathis: Could you elaborate? I'm having trouble remembering what you're referring to.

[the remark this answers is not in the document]
sed-i: I was referring to [this](https://github.com/canonical/catalogue-k8s-operator/blob/5d1de5175799b416bb3ea554945af7f6b75f7a8e/.github/workflows/pull-request.yaml#L14):

```yaml
    with:
      charm-path: "charm"
```

But I realize it's not relevant in this case.

So what's breaking without this stub?

sed-i (approved):

metadata.yaml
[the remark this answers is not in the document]
dstathis: IIRC the harness would not work without it.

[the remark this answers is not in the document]
sed-i: `tox -e unit` is setting it up though `cp k8s_metadata.yaml metadata.yaml`.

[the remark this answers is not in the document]
dstathis: Looks like it isn't needed anymore. Removed.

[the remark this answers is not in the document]
dstathis: Okay looks like it was for the library check

PR #67, opened 2022-12-13
update libraries

lucabello:


sed-i:
k8s service patch v0 is no longer in the o11y libs repo.
[Recently](https://github.com/canonical/observability-libs/pull/34) I ran `no_implicit_optional` there so one way to "fix" the static check is switching here from v0 to v1. [Example](https://github.com/canonical/catalogue-k8s-operator/blob/7b057747de5d132e290f52df957b8648a402c3d4/charm/src/charm.py#L44).

dstathis (approved):

PR #82, opened 2023-02-01
Use a ROCK image

rbarry82:
## Issue
Move to the `grafana-agent` ROCK image on GHCR

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -1,8 +1,6 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-# Stub file to make CI work correctly
-
 name: grafana-agent-k8s
 
 description: |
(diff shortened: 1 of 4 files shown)


simskij (approved):

metadata.yaml
```diff
@@ -11,3 +9,40 @@ summary: |
   Grafana Agent is a telemetry collector for sending metrics, logs, and trace data to the opinionated Grafana observability stack.
 maintainers:
     - Jose Massón <jose.masson@canonical.com>
+
```
PietroPasotti, line 12: what do these changes have to do with using a rock image? Am I missing something?

rbarry82: These slipped through on the git commit without notice until now, it seems like. Since the [`tox` "render" simply overwrites this file in its entirety](https://github.com/canonical/grafana-agent-k8s-operator/blob/main/tox.ini#L37) it's probably harmless, but I'll put it back out on the next patch

PR #83, opened 2023-02-05
Fix release

dstathis:

### DIFF
--- a/.github/workflows/release-edge.yaml
+++ b/.github/workflows/release-edge.yaml
@@ -9,8 +9,26 @@ jobs:
   render:
     runs-on: ubuntu-20.04
     steps:
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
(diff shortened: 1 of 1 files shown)


.github/workflows/release-edge.yaml
```diff
@@ -9,8 +9,26 @@ jobs:
...
+          name: metadata
+          path: metadata.yaml
+      - name: Upload charm.py
+        uses: actions/upload-artifact@v3
+        with:
+          name: charm_py
+          path: src/charm.py
   release:
-    uses: canonical/observability/.github/workflows/release-charm.yaml@main
+    uses: dstathis/observability/.github/workflows/release-charm.yaml@artifacts
```
simskij, line 31: *cough* this line should probably be reverted?

dstathis: This PR is completely wrong I believe. Closing and reopening the correct one.

PR #84, opened 2023-02-10
Fix release CI

### LINKED ISSUE #75: The "Release" workflow fails because no entrypoint is rendered
### Bug Description

The release workflow [recently failed](https://github.com/canonical/grafana-agent-k8s-operator/actions/runs/3990598358/jobs/6845258863) because at that point there is no `charm.py` rendered in `src`.

### To Reproduce

Simply run `charmcraft pack` without `tox -e render-k8s` first.

### Environment

Not relevant.

### Relevant log output

```shell
$ charmcraft pack
Packing the charm
Lint Errors:
- entrypoint: Cannot find the entrypoint file: '/root/prime/src/charm.py' (https://juju.is/docs/sdk/charmcraft-analyzers-and-linters#heading--entrypoint)
Aborting due to lint errors (use --force to override).
Failed to build charm for bases index '0'.
```

### Additional context

_No response_

dstathis:
## Issue
Fixes #75

## Solution
Render additional build files and upload them as an artifact to the observability lib.

## Testing Instructions
Look for green!

## Release Notes
-

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -106,7 +106,8 @@ description = Run integration tests
 deps =
     aiohttp
     asyncstdlib
-    juju
+    # Libjuju needs to track the juju version
+    juju ~= 3.0.0
     pytest
     prometheus-api-client
     pytest-operator
--- a/.github/workflows/release-edge.yaml
+++ b/.github/workflows/release-edge.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Render k8s Charm
         run: tox --workdir $GITHUB_WORKSPACE/main -e render-k8s
   release:
-    uses: dstathis/observability/.github/workflows/release-charm.yaml@main
+    uses: canonical/observability/.github/workflows/release-charm.yaml@main
     needs: render
     secrets: inherit
     with:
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -106,7 +106,8 @@ description = Run integration tests
 deps =
     aiohttp
     asyncstdlib
-    juju
+    # Libjuju needs to track the juju version
+    juju ~= 3.0.0
```
rbarry82, line 110:
```suggestion
    juju > 3.1
```

dstathis: This doesn't work. You need libjuju 3.0.* for juju 3.0.*

rbarry82: `<3.1` is also libjuju 3.0.4. [There aren't other options](https://pypi.org/project/juju/#history). It's functionally exactly the same as `~= 3.0.0`, just without syntax which confuses anyone who has ever used `~=` in JS/Ruby/Perl/shell, where `=~ || ~=` is generally a regexp matcher, including in our own PromQL/LogQL.

It's not a requirement and more of a nit because "`~=` means to stay in this semver series" is a bizarre, requirements.txt-only thing

simskij (approved):

PR #89, opened 2023-02-27
Machine charm bugfixes that enable the charm to run without errors. T…

awnns:
…he following were fixed.

1. Due to the ordering of life cycle events, we have to wait until the relation_joined event of the juju-info relation before we try to read the unit from the relation. The is_ready method is updated to reflect this.
2. The old method for fetching the unit changes the state of the charm instance. This is problematic because there are event handlers in the charm that perform this operation twice. A new property was added called principal_unit which fetches the unit in a safe but ugly manner.
3. There were a few instances of the dict.update method being used to merge dictionaries. The problem here is that dict.update updates dictionaries in place and does not return the new value. In these particular cases, dictionary unpacking is better.
4. A return value was added to GrafanaAgentCharm._loki_config.
5. Fixed some typos.

The grafana-agent itself still doesn't work because the configuration files produced aren't quite right.

### DIFF
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -402,7 +402,7 @@ class GrafanaAgentCharm(CharmBase):
                 }
             )
 
-        if self.is_machine:
+        if self.is_machine():
             logs_config["logs"]["configs"].append(
                 {
                     "name": "log_file_scraper",
@@ -414,13 +414,16 @@ class GrafanaAgentCharm(CharmBase):
                                 "labels": {
                                     "__path__": "/var/log/*",
                                     "__path_exclude__": "/var/log/positions.yaml",
-                                }.update(self.get_principal_labels())
+                                    **self.get_principal_labels(),
+                                }
                             },
                         },
                         {"job_name": "syslog", "journal": {"labels": self.get_principal_labels()}},
                     ],
                 }
             )
+            
+        return logs_config
 
     def _reload_config(self, attempts: int = 10) -> None:
         """Reload the config file.
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -111,6 +111,7 @@ class GrafanaAgentCharm(CharmBase):
         """Gets the raw output from `agent -version`."""
         raise NotImplementedError("Please override the agent_version_output method")
 
+    @property
     def is_ready(self):
         """Checks if the charm is ready for configuration."""
         raise NotImplementedError("Please override the is_ready method")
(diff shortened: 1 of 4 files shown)


rbarry82:
Ugh, github. Can you rebase this, please @awnns ? I pushed cleanup [here](https://github.com/canonical/grafana-agent-k8s-operator/commit/87bf25cff2c1c3f0c7670a442e0a8271d5d69713), but it's not showing in your PR

src/machine_charm.py
```diff
@@ -93,7 +93,7 @@ def on_remove(self, _) -> None:
 
     def is_ready(self):
         """Checks if the charm is ready for configuration."""
-        return self._is_installed()
+        return self._is_installed() and self.principal_unit
```
PietroPasotti, line 96: a bit unclear why there being a principal unit means that the charm is ready. 
Could you add a comment?

src/grafana_agent.py
```diff
@@ -402,7 +402,7 @@ def _loki_config(self) -> dict:
                 }
             )
 
-        if self.is_machine:
+        if self.is_machine():
```
PietroPasotti, line 405: Minor stylistic preference (but also consistency with other similar methods in other o11y charms): I'd fix this by making `is_machine` a property.

src/machine_charm.py
```diff
@@ -140,29 +140,44 @@ def _get_principal_relation(self) -> Optional[Relation]:
             return self.model.relations["juju-info"][0]
         else:
             return None
-
+        
+    @property
+    def principal_unit(self):
```
PietroPasotti, line 145: instead of a property, I'd make this an instance attribute. Set it in `__init__` and be done with it, no? it's not going to change later. Also `is_ready`, by the way.

PietroPasotti (commented):
A few nits, didn't test the charm yet (we're swamped this sprint).

src/machine_charm.py
awnns: I will get to it later, but the gist of it is that the is_ready method is used to detect when to actually start writing the configuration files. We need to know the name of the principal unit because it is used to generate the configuration file. I suppose we could also just adapt the config file generators, but at present there are a number of interconnected methods that will throw exceptions if we don't have a principal unit name. The name of the principal unit has to be fetched from the juju-info relation (which is what makes the subordinate inject into the principal unit). However, the field in the relation is only populated on the relation_joined event so we can not use it in the setup phase. There are event handlers that use the principal unit name that are run in the setup phase, so it is guaranteed to fail if we don't check for the presence of the principal unit. I will write something more condensed.

dstathis:
I think these changes have already been merged in to the machine_charm branch. Feel free to reopen if needed.

PR #90, opened 2023-02-28
Cleanup etc.

sed-i:
Part cleanup, part ideas from https://git.launchpad.net/charm-local-users/tree/.


charmcraft.yaml
```diff
@@ -5,10 +5,15 @@ type: charm
...
-      channel: "20.04"
+      channel: "22.04"
+      architectures: ["amd64"]
     run-on:
+    - name: "ubuntu"
+      channel: "22.04"
+      architectures: ["amd64"]
     - name: "ubuntu"
       channel: "20.04"
+      architectures: ["amd64"]
```
sed-i, line 16: Not sure if this 20.04 section is needed.

machine_metadata.yaml
```diff
@@ -15,7 +15,7 @@ website: https://charmhub.io/grafana-agent
 source: https://github.com/canonical/grafana-agent-k8s-operator
 issues: https://github.com/canonical/grafana-agent-k8s-operator/issues
 
-series: []
+series: [jammy]
```
sed-i, line 18: This was needed to "fix" this:
```
ERROR cannot add relation "agent:metrics-endpoint kafka:metrics-endpoint": principal and subordinate applications' series must match
```

metadata.yaml
```diff
@@ -1,48 +0,0 @@
-# Copyright 2021 Canonical Ltd.
```
sed-i, line 1: This file is already in .gitignore.

machine_metadata.yaml
rbarry82: We probably shouldn't be relating there anyway, but this is a whole new world for us. It [doesn't look that particular](https://github.com/juju/juju/blob/develop/state/state.go#L2232). I think we may actually be able to expand this to "any series which this charm will work as a subordinate of."

Considering that we aren't actually messing with dpkg/packages at all, or compiled code (which would be sort-of fine in a snap anyway), I suspect it's reasonable to have `[focal, jammy]`, at least.

charmcraft.yaml
rbarry82: It is.

Abuelodelanada (approved):

rbarry82 (approved):

PR #92, opened 2023-03-01
Class split

rbarry82:
## Issue
Make the codebase less muddy.

## Solution
Move machine agent stuff (`*_primary`) out of the main class, sync up config locations/generation so it's not a series of appended dicts from different places. 

Create methods to be overridden for per-machine/k8s config block additions (this also makes the `is_machine` method unnecessary), which are appended as part of the "wholesale" config generation in the main class body.

Fix the unit tests.

### DIFF
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -48,7 +48,7 @@ class GrafanaAgentCharm(CharmBase):
     """Grafana Agent Charm."""
 
     _name = "agent"
-    _promtail_positions = "/tmp/positions.yaml"
+    _promtail_positions = "/run/promtail-positions.yaml"
     _http_listen_port = 3500
     _grpc_listen_port = 3600
 
@@ -195,7 +195,11 @@ class GrafanaAgentCharm(CharmBase):
         self._update_status()
 
     def _update_status(self) -> bool:
-        """Update the status to reflect the status quo."""
+        """Update the status to reflect the status quo.
+
+        Returns:
+            True if the status was set to Active; False otherwise.
+        """
         if len(self.model.relations["metrics-endpoint"]):
             if not len(self.model.relations[REMOTE_WRITE_RELATION_NAME]):
                 self.unit.status = WaitingStatus("no related Prometheus remote-write")
@@ -226,6 +230,7 @@ class GrafanaAgentCharm(CharmBase):
 
         if config == old_config:
             # Nothing changed, possibly new install. Set us active and move on.
+            self.unit.status = ActiveStatus()
             return
 
         try:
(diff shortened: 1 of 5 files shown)


src/machine_charm.py
```diff
@@ -138,17 +131,56 @@ def restart(self) -> None:
...
+            {
+                "name": "log_file_scraper",
+                "clients": self._loki_consumer.loki_endpoints,
+                "scrape_configs": [
+                    {
+                        "job_name": "varlog",
+                        "static_configs": {
+                            "labels": {
+                                "__path__": "/var/log/*",
+                                "__path_exclude__": "/var/log/positions.yaml",
```
simskij, line 174: positions are stored in `/tmp/positions.yaml`, so should not be needed.

src/machine_charm.py
```diff
@@ -138,17 +131,56 @@ def restart(self) -> None:
...
+        return [
+            {
+                "name": "log_file_scraper",
+                "clients": self._loki_consumer.loki_endpoints,
+                "scrape_configs": [
+                    {
+                        "job_name": "varlog",
+                        "static_configs": {
+                            "labels": {
+                                "__path__": "/var/log/*",
```
simskij, line 173: we should probably change this to avoid trying to tail zip files and rotated logs.

```suggestion
                                "__path__": "/var/log/*log",
```

simskij (approved):
See my separate comments. Once addressed, I think this can be merged.

src/grafana_agent.py
```diff
@@ -390,6 +345,7 @@ def _loki_config(self) -> dict:
                 {
                     "name": "push_api_server",
                     "clients": self._loki_consumer.loki_endpoints,
+                    "positions": {"filename": "/tmp/positions.yaml"},
```
sed-i, line 348: What's `/tmp/positions.yaml`?
Should we use `tempfile` methods around this?

dstathis (approved):

src/grafana_agent.py
```diff
@@ -282,102 +281,58 @@ def _integrations_config(self) -> dict:
...
-            },
-            {
-                "source_labels": ["__address__"],
-                "target_label": "juju_unit",
-                "replacement": juju_unit,
-            },
-        ]
 
         conf = {
-            "integrations": {
```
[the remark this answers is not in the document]
rbarry82: It [comes from here instead](https://github.com/canonical/grafana-agent-k8s-operator/pull/92/files#diff-8852e3cdd024457a879f382cc1cc1bfed3d52709253ad614e1830d443cd9bc8cR259).

PR #96, opened 2023-03-02
Fix indentation

rbarry82:
## Issue
Make the static configs into a list so it parses

### DIFF
--- a/src/machine_charm.py
+++ b/src/machine_charm.py
@@ -154,13 +154,15 @@ class GrafanaAgentK8sCharm(GrafanaAgentCharm):
                 "scrape_configs": [
                     {
                         "job_name": "varlog",
-                        "static_configs": {
-                            "targets": ["localhost"],
-                            "labels": {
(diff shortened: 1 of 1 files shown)


src/machine_charm.py
```diff
@@ -154,13 +154,15 @@ def _additional_log_configs(self) -> List[Dict[str, Any]]:
...
-                            "labels": {
-                                "__path__": "/var/log/*log",
-                                **self._principal_labels,
-                            },
-                        },
+                        "static_configs": [
+                            {
+                                "targets": ["localhost"],
+                                "labels": {
+                                    "__path__": "/var/log/*log",
```
sed-i, line 161: Do we need a period there?
```suggestion
                                    "__path__": "/var/log/*.log",
```

sed-i (approved):

src/machine_charm.py
sed-i: Syslog.. scratch that :)

PR #98, opened 2023-03-02
Clear machine metadata relations

sed-i:
## Issue
Afaiu, in order to remove the two relations in machine's metadata that `cos_machine` is intended to replace, we would need to further separate between the machine and k8s codes.

## Solution
Remove from machine metadata:
- prometheus_scrape requirer
- loki_push_api provider

### DIFF
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -11,6 +11,7 @@ from collections import namedtuple
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import yaml
+from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.loki_k8s.v0.loki_push_api import LokiPushApiConsumer
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
     PrometheusRemoteWriteConsumer,
@@ -84,6 +85,15 @@ class GrafanaAgentCharm(CharmBase):
 
         self._loki_consumer = LokiPushApiConsumer(
             self, relation_name="logging-consumer", alert_rules_path=self.loki_rules_paths.dest
+        )
+
+        self._grafana_dashboards_provider = GrafanaDashboardProvider(
+            self,
+            relation_name="grafana-dashboards-provider",
+            dashboards_path=self.dashboard_paths.dest,
+        )
+        self.framework.observe(
+            self._grafana_dashboards_provider.on.dashboard_status_changed, self._dashboards_changed
         )
 
         self.framework.observe(self.on.upgrade_charm, self._update_metrics_alerts)
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -11,11 +11,10 @@ from collections import namedtuple
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import yaml
-from charms.loki_k8s.v0.loki_push_api import LokiPushApiConsumer, LokiPushApiProvider
+from charms.loki_k8s.v0.loki_push_api import LokiPushApiConsumer
 from charms.prometheus_k8s.v0.prometheus_remote_write import (
     PrometheusRemoteWriteConsumer,
 )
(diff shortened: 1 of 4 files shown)


src/grafana_agent.py
```diff
@@ -147,22 +134,6 @@ def _additional_log_configs(self) -> List[Dict[str, Any]]:
...
-        )
-
-    def _loki_alerts(self, event):
-        self.update_alerts_rules(
-            event,
-            alerts_func=self._loki_provider.alerts,
-            reload_func=self._loki_consumer._reinitialize_alert_rules,
-            mapping=self.loki_rules_paths,
-        )
-
```
rbarry82, line 165: These need to stay in the base class.

The comments in the dailies about having remapped attributes with some more... generic name than `self._scrape` or `self._loki_provider` which can be overridden or remapped in child classes to fetch all alerts from the "unified" relation (by type) *or* to get them from the k8s libraries here is the primary complex logic in Grafana Agent, and the reason why I was against splitting the repos.

We'd additionally have one for dashboards (probably after renaming `update_alert_rules` to `update_data` or something with the added type). This "shuffling" of data from one bag to another one is really the reason why they're merged, other than the config being the same.

src/machine_charm.py
```diff
@@ -47,6 +47,12 @@ def __init__(self, *args):
         self.framework.observe(self.on.remove, self.on_remove)
         self.framework.observe(self.on["juju_info"].relation_joined, self._update_config)
 
+        # TODO observe cos_machine changed and call e.g.
```
rbarry82, line 50: In the end, the machine charm will be able to just call `self._update_config(event)`, basically. 

The logic in `_update_status` is somewhat too dependent on the behavior of the k8s version, since the machine charm probably _always_ wants a remote write but doesn't care about a metrics endpoint, but that's for a different patch.

Either way, both log target changes and scrape changes will want to call `update_config`, so may as well just clal that endpoint for every machine event.

sed-i:
I pulled your changes @dstathis into here. cc: @rbarry82.

sed-i:
Deployed and related to kafka over juju-info. No errors.

dstathis (approved):

src/k8s_charm.py
```diff
@@ -68,6 +81,18 @@ def on_pebble_ready(self, _) -> None:
             )
         self._update_status()
 
+    def metrics_rules(self):
```
PietroPasotti, line 84: why not make this an instance attribute set in __init__?
```
        self._scrape = MetricsEndpointConsumer(self)
        self.metrics_rules=self._scrape.alerts
```

PietroPasotti: feels like a lot of code to proxy a few methods.

src/machine_charm.py
```diff
@@ -80,6 +80,18 @@ def on_remove(self, _) -> None:
         if self._is_installed:
             raise GrafanaAgentInstallError("Failed to uninstall grafana-agent")
 
+    def metrics_rules(self) -> list:
+        """Return a list of metrics rules."""
+        return []
```
PietroPasotti, line 85: is this a `#TODO` or is this a method that has no meaning in the context of the machine charm, and if so, should it ever be called?

PietroPasotti (changes requested):
main blocker is clarity on the role of the methods added to the machine charm

src/machine_charm.py
dstathis: It is a todo

src/k8s_charm.py
dstathis: I think it is a bit more clear what is going on when you have virtual methods in the base class. I'm not completely apposed to this approach though

PietroPasotti (approved):

src/k8s_charm.py
PietroPasotti: to be discussed as a team and eventually addressed in a separate PR, don't consider this a blocker as it's purely aesthetic

PR #113, opened 2023-03-07
Use new Juju for CI

rbarry82:
## Issue
Use 3.1/edge so CI passed. Update-lib


sed-i:
Revision 26 of charm 'grafana-agent-k8s' released to edge (attaching resources: 'agent-image' r21)

PR #115, opened 2023-03-08
More tests

PietroPasotti:
added some more scenario tests verifying behaviour with cos-machine and juju-info:
- either one
- none
- both


lucabello (approved):

tests/scenario/test_machine_charm/test_relation_priority.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@pytest.fixture
+def mock_cfg_path(tmp_path):
+    return tmp_path / "foo.yaml"
+
+
+@pytest.fixture(autouse=True)
+def patch_all(mock_cfg_path):
+    grafana_agent.CONFIG_PATH = mock_cfg_path
```
sed-i, line 35: This is a monkeypatch.
Did you mean: `patcher = patch('grafana_agent.CONFIG_PATH', mock_cfg_path)`?

tests/scenario/test_machine_charm/test_relation_priority.py
```diff
@@ -0,0 +1,158 @@
...
+                "dashboards": {
+                    "dashboards": [
+                        COSMachineProvider._encode_dashboard_content("very long dashboard")
+                    ],
+                },
+            }
+        )
+    }
+    trigger(
+        "start",
```
sed-i, line 148: All the tests in the file only trigger start. Does this make the tests brittle?
I.e. these seem borderline whitebox tests.

tests/scenario/test_machine_charm/test_relation_priority.py
```diff
@@ -0,0 +1,158 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+import json
+from unittest.mock import patch
+
+import pytest
+from charms.grafana_agent.v0.cos_machine import COSMachineProvider
+from scenario import Relation, State
+from scenario import trigger as _trigger
```
sed-i, line 9: How are you able to use `trigger` below if you import it as `_trigger`?

PietroPasotti: see line 17

tests/scenario/test_machine_charm/test_relation_priority.py
PietroPasotti: Good point, however, what I'm testing here is not the output state but the charm state, as you can tell by the `post_event` callbacks. And also I'm not interested in why the charm is executed, I just want it to initialize to then verify the return values of some of its properties and methods.

So it is brittle in that if we ever stop listening to 'start', we'll have to use some other event to have the charm execute. But it's not brittle in that we're not in fact checking anything specific to the start event in itself. Any other event will do. 
We could parametrize and run these tests for a bunch other events as well (or all of them), but it felt pointless.

tests/scenario/test_machine_charm/test_relation_priority.py
PietroPasotti: is that better?

tests/scenario/test_machine_charm/test_relation_priority.py
PietroPasotti: You're right in that the box here is much whiter than by other scenario tests I've written, but it felt a bit premature to run assertions on the output config given that it's still fluid

tests/scenario/test_machine_charm/test_relation_priority.py
sed-i: Less unexpected side-effects, although here in tests this is unlikely to matter.

tests/scenario/test_machine_charm/test_relation_priority.py
sed-i: Not a priority!

lucabello:
The static-lib check is not failing for the **cos-agent** library (only warnings for that), but for **cos-machine** which I think is some old code leftover that somehow creeped in your PR :)

PietroPasotti:
> The static-lib check is not failing for the **cos-agent** library (only warnings for that), but for **cos-machine** which I think is some old code leftover that somehow creeped in your PR :)
> 
> Deleting that should be enough!

Ha, you're right!
Must have missed it in the merge

lucabello (approved):

PR #116, opened 2023-03-08
Fix position files

### LINKED ISSUE #109: (machine_charm) When related over loki_push_api, grafana-agent fails to start
### Bug Description

After relating to loki, the grafana-agent service fails to start.

### To Reproduce

1. Deploy kafka or zookeeper
2. Deploy grafana-agent and relate to the above charm
3. Relate to loki via cross-model relation
4. `systemctl status snap.grafana-agent.grafana-agent.service`

### Environment

-

### Relevant log output

```shell
ubuntu@juju-488a58-4:~$ /usr/bin/snap run grafana-agent
2023/03/07 15:09:25 error loading config file /etc/grafana-agent.yaml: Loki configs push_api_server and log_file_scraper must have different positions file paths
```

### Additional context

_No response_

dstathis:
## Issue
Fixes #109

## Solution
Use different names for the positions files

## Testing Instructions
Relate over loki_push_api and check `systemctl status snap.grafana-agent.grafana-agent.service`

## Release Notes
Fixed an issue where Grafana Agent could not start if related to loki.


rbarry82 (approved):
LGTM.

Amazing that every different scraper needs its own YAML instead of just making a new key.

Obviously the linter, again...

Abuelodelanada (approved):
Shouldn't we put these in constants, so it is easy to modify in the future?

rbarry82:
For me, that's six of one and half a dozen of another. The only place the constant would be used is here, and it's swapping one hardcoded string for another.

If Grafana Agent isn't able to share, I wonder instead if we should make the name more dynamic, so it's something like:

```python
name = # generate_some_meaningful_identifier
{
  "name": name,
  "positions": { "filename": f"/run/{name}_positions.yaml" },
  ..
}
```

To future proof a little against the possibility of some other possible need for a position file coming up in the future and us forgetting that bad things happen if they share.

awnns:
I played around a little with this. You can also set a folder in which the grafana-agent will generate its own position files based on the job name so you don't have to set the name manually for every job. There's a field on the same level as "configs" called "positions_directory" (see https://grafana.com/docs/agent/latest/configuration/logs-config/). I had to put it in /var/snap/[...] to get it to work permissions wise, but that did the trick for me at least. Have you considered it?

rbarry82:
I don't know if any of us came across that in the docs, but it's a great find we should definitely use. Thanks @awnns !

PR #121, opened 2023-03-08
Use cosl for Rules and JujuTopology

rbarry82:
## Issue
Swap to COSL instead of re-using the Prom/Loki AlertRules to minimize deps for end-users and side effects

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
+cosl
 ops
 requests
 kubernetes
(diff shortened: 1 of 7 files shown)


Abuelodelanada (approved):

requirements.txt
```diff
@@ -1,5 +1,6 @@
 # Copyright 2021 Canonical Ltd.
 # See LICENSE file for licensing details.
+cosl
```
sed-i, line 3: Might as well add it to PYDEPS so we don't forget when we remove it from here.

sed-i (approved):

requirements.txt
rbarry82: It's already there :)

PR #122, opened 2023-03-09
top level documentation for cos_machine lib 

Abuelodelanada:
## Issue

Top level documentation for cos_machine lib

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_machine.py
+++ b/lib/charms/grafana_agent/v0/cos_machine.py
@@ -56,7 +56,7 @@ The constructor of `COSMachineProvider` has only one required and eight optional
 - `recurse_rules_dirs`: This paramenters set wheter Grafana Agent machine Charmed Operator has to search
     alert rules files recursively in the previous two directories or not.
 
-- `logs_slots`: Snap slots to connect to for scraping logs in the form ["snap-name:slot", ...].
+- `log_slots`: Snap slots to connect to for scraping logs in the form ["snap-name:slot", ...].
 
 - `dashboard_dirs`: List of directories where the dashboards are stored in the Charmed Operator.
 
@@ -97,7 +97,7 @@ class TelemetryProviderCharm(CharmBase):
             metrics_rules_dir="./src/alert_rules/prometheus",
             logs_rules_dir="./src/alert_rules/loki",
             recursive_rules_dir=True,
-            logs_slots=["my-app:slot"],
+            log_slots=["my-app:slot"],
             dashboard_dirs=["./src/dashboards_1", "./src/dashboards_2"],
             refresh_events=["update-status", "upgrade-charm"],
         )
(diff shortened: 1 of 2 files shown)


lib/charms/grafana_agent/v0/cos_machine.py
```diff
@@ -1,8 +1,166 @@
...
+```python
+    def __init__(
+        self,
+        charm: CharmType,
+        relation_name: str = DEFAULT_RELATION_NAME,
+        metrics_endpoints: Optional[List[dict]] = None,
+        metrics_rules_dir: str = "./src/prometheus_alert_rules",
+        logs_rules_dir: str = "./src/loki_alert_rules",
+        recurse_rules_dirs: bool = False,
+        logs_slots: Optional[List[str]] = None,
```
sed-i, line 40: Slightly out of scope:
```suggestion
        log_slots: Optional[List[str]] = None,
```

lib/charms/grafana_agent/v0/cos_machine.py
```diff
@@ -1,8 +1,166 @@
...
+                {"path": "/metrics", "port": 9000},
+                {"path": "/metrics", "port": 9001},
+                {"path": "/metrics", "port": 9002},
+            ],
+            metrics_rules_dir="./src/alert_rules/prometheus",
+            logs_rules_dir="./src/alert_rules/loki",
+            recursive_rules_dir=True,
+            logs_slots=["my-app:slot"],
+            dashboard_dirs=["./src/dashboards_1", "./src/dashboards_2"],
+            refresh_events=["update-status", "upgrade-charm"],
```
sed-i, line 106: Odd that the library doesn't already observe `upgrade-charm` on behalf of the charm. We should probably add it in and drop it from the example here.

sed-i (approved):
This doc addition certainly helped me understand better what is going on!

lib/charms/grafana_agent/v0/cos_machine.py
Abuelodelanada: `upgrade-charm` is certainly a bad example here, since [this event will end up triggering a `start` event](https://juju.is/docs/sdk/upgrade-charm-event#heading--emission-sequence) which we observe on behalf of the charm.

rbarry82: Assuming that we trust the workflow, `config-changed` is at least a safe default (and also in that emission sequence).

I actually wonder whether we shouldn't make this example smaller. Showing an example with every possible default arg overridden is great for API docs, but the signature and docstrings can help there. This sort of reads like every user _must_ provide all of these fields, when almost none of them are strictly required as long as the defaults are ok.

rbarry82 (commented):
Minor UX comment, but this looks great!

lib/charms/grafana_agent/v0/cos_machine.py
sed-i: - config-changed is currently observed only if `refresh_events` is None.
- I do not see any observer for `start`

Given the above I think we need to always observe config-changed. And document here that there is no need to observe either config-changed or upgrade-charm.

Abuelodelanada: @sed-i sorry I got confused, the `start` event is observer in the machine_charm.py file

Abuelodelanada: Let's follow this on [this issue](https://github.com/canonical/grafana-agent-k8s-operator/issues/126) and merge this PR

PR #125, opened 2023-03-10
Loki shared logs monitoring

PietroPasotti:
adds separate scrape configs for each connected snap (todo: verify that it works like I think it should)

and a big scenario test for the resulting config.


lucabello (approved):

lib/charms/grafana_agent/v0/cos_machine.py
```diff
@@ -255,13 +257,29 @@ def metrics_jobs(self) -> List[Dict]:
...
+                        )
+                    else:
+                        plugs.append(target)
+
+        endpoints = []
+        for plug in plugs:
+            if ":" not in plug:
+                logger.error(f"invalid plug definition received: {plug}. Ignoring...")
+            else:
+                endpoint = SnapEndpoint(*plug.split(":"))
```
rbarry82, line 280: 💯 `NamedTuple` is the best

rbarry82 (approved):

PR #129, opened 2023-03-14
Rename cos-machine to cos-agent

sed-i:
## Issue
Lib/iface name has "machine" in it, which is redundant.

## Solution
Rename to `cos-agent`.

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -3,26 +3,26 @@
 
 r"""## Overview.
 
-This library can be used to manage the cos_machine relation interface:
+This library can be used to manage the cos_agent relation interface:
 
-- `COSMachineProvider`: Use in machine charms that need to have a workload's metrics
+- `COSAgentProvider`: Use in machine charms that need to have a workload's metrics
   or logs scraped, or forward rule files or dashboards to Prometheus, Loki or Grafana through
   the Grafana Agent machine charm.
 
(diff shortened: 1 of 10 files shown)


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -3,26 +3,26 @@
...
+This library can be used to manage the cos_agent relation interface:
 
-- `COSMachineProvider`: Use in machine charms that need to have a workload's metrics
+- `COSAgentProvider`: Use in machine charms that need to have a workload's metrics
   or logs scraped, or forward rule files or dashboards to Prometheus, Loki or Grafana through
   the Grafana Agent machine charm.
 
-- `COSMachineConsumer`: Used in the Grafana Agent machine charm to manage the requrier side of 
-  the `cos_machine` interface.
+- `COSAgentConsumer`: Used in the Grafana Agent machine charm to manage the requrier side of
```
rbarry82, line 12:
```suggestion
- `COSAgentConsumer`: Used in the Grafana Agent machine charm to manage the requirer side of
```

rbarry82 (approved):
Typo. 

I'm not sold on the idea that having `machine` in the name is redundant (which is only true assuming that one already knows what the intended purpose is), nor that this interface would/should never be used for any other "combined" interface (especially as they move towards [Flow](https://grafana.com/docs/agent/latest/flow/), which would need a somewhat careful evaluation vis-a-vis Vector) but I'm in

PR #132, opened 2023-03-15
Reconcile CI

rbarry82:
## Issue
Fetch all libs, pull in the uk8s_group helper, xfail likely failures (like trying to upgrade from CH when the version in CH will have failing patches ATM)


lib/charms/prometheus_k8s/v0/prometheus_scrape.py
```diff
@@ -2195,6 +2259,9 @@ def set_alert_rule_data(self, name: str, unit_rules: dict, label_rules: bool = T
         The unit rules should be a dict, which is has additional Juju topology labels added. For
```
simskij, line 2259:
```suggestion
        The unit rules should be a dict, which has additional Juju topology labels added. For
```

Drive-by change. Not really part of the PR scope. Do with it what you want. :)

simskij (approved):

Issue #107, opened 2023-03-07
Are we really going to have more than 1 layer of closures nested?


PietroPasotti:
Harmonize this in `prometheus_scrape` so we don't need the horrible `_recurse_call_chain` thing.

_Originally posted by @PietroPasotti in https://github.com/canonical/grafana-agent-k8s-operator/pull/95#discussion_r1127761141_

rbarry82:
Actually [done](https://github.com/canonical/prometheus-k8s-operator/pull/465)! It just needs a rebuild of `grafana-agent-k8s` with the new lib.

PR #145, opened 2023-03-23
Be looser with finding snap plug mounts

rbarry82:
If there's only one matching entry with an owning snap in fstab, use that without checking the source at all.

If there's more than one, coalesce the relative path and the provided endpoint name into sets of unique characters, then find the one with the longest intersection as the most likely candidate.

Add a couple more assertions in the existing unit test to verify this.

### DIFF
--- a/src/machine_charm.py
+++ b/src/machine_charm.py
@@ -88,11 +88,27 @@ class SnapFstab:
 
         self.entries = entries
 
-    def entry(self, owner: str, endpoint_name: str) -> Optional[_SnapFstabEntry]:
+    def entry(self, owner: str, endpoint_name: Optional[str]) -> Optional[_SnapFstabEntry]:
         """Find and return a specific entry if it exists."""
-        entries = [
-            e for e in self.entries if e.owner == owner and endpoint_name in e.endpoint_source
-        ]
+        entries = [e for e in self.entries if e.owner == owner]
+
+        if len(entries) > 1 and endpoint_name:
+            # If there's more than one entry, the endpoint name may not directlly map to
+            # the sourcce *or* path. charmed-kafka uses 'logs' as the plug name, and maps
+            # .../common/logs to .../log inside Grafana Agent
+            #
+            # In this case, for a cheap comparison (rather than implementing some recursive
+            # LCS just for this, convert all possible endpoint sources into a list of unique
+            # characters, as well as the endpoint name, and build a dict of entries with
+            # a value that's the length of the intersection, the pick the first one.
+            character_matches = {
+                e: len(set(endpoint_name) & set(e.endpoint_source)) for e in entries
+            }
+            sorted_matches = {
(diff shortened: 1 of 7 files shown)


src/machine_charm.py
```diff
@@ -88,11 +88,27 @@ def __post_init__(self):
...
+
+        if len(entries) > 1 and endpoint_name:
+            # If there's more than one entry, the endpoint name may not directlly map to
+            # the sourcce *or* path. charmed-kafka uses 'logs' as the plug name, and maps
+            # .../common/logs to .../log inside Grafana Agent
+            #
+            # In this case, for a cheap comparison (rather than implementing some recursive
+            # LCS just for this, convert all possible endpoint sources into a list of unique
+            # characters, as well as the endpoint name, and build a dict of entries with
+            # a value that's the length of the intersection, the pick the first one.
```
sed-i, line 103: Could you add a "why" to the docstring? I don't understand why we're doing this.

rbarry82:
**NOTE** practically all of the changes in the libs are for the "other side" of the relation and already current with what we're testing. This is just for coherency, and because it's needed for #107 for a simple `@property` decorator in Prometheus

src/machine_charm.py
rbarry82: Done.

sed-i:
I think this change may have broken some tests
```
FAILED tests/unit/k8s/test_alerts.py::TestPrometheusRules::test_consumes_prometheus_rules - charms.observability_libs.v0.juju_topology.InvalidUUIDError: 'f2c1b2a6-e006-11eb-ba80...
FAILED tests/unit/k8s/test_alerts.py::TestPrometheusRules::test_forwards_prometheus_rules - TypeError: 'dict' object is not callable
FAILED tests/unit/k8s/test_alerts.py::TestLokiRules::test_consumes_loki_rules - charms.observability_libs.v0.juju_topology.InvalidUUIDError: 'f2c1b2a6-e006-11eb-ba80...
FAILED tests/unit/k8s/test_alerts.py::TestLokiRules::test_forwards_loki_rules - charms.observability_libs.v0.juju_topology.InvalidUUIDError: 'f2c1b2a6-e006-11eb-ba80...
SUBFAIL tests/unit/k8s/test_relation_status.py::TestRelationStatus::test_with_relations - TypeError: 'dict' object is not callable
SUBFAIL tests/unit/k8s/test_relation_status.py::TestRelationStatus::test_with_relations - AssertionError: BlockedStatus("Missing relation: 'send-remote-write'") is not an inst...
FAILED tests/unit/k8s/test_scrape_configuration.py::TestScrapeConfiguration::test_remote_write_configuration - TypeError: 'dict' object is not callable
FAILED tests/unit/k8s/test_scrape_configuration.py::TestScrapeConfiguration::test_scrape_without_remote_write_configuration - TypeError: 'dict' object is not callable
```

PR #149, opened 2023-03-23
Typed metrics endpoint dict

### LINKED ISSUE #106: add typing for endpoint
```
if typing.TYPE_CHECKING:
    from typing import TypedDict  # if the smallest python version we support has it! Otherwise, you could wrap this all in a try/except and if there's an error define `_EndpointDict = dict` and that's it.
    
    class _EndpointDict(TypedDict):  
        path:str
         port:int # ?
```

```diff
-        metrics_endpoints: List[dict] = [DEFAULT_METRICS_ENDPOINT],
+        metrics_endpoints: List["_EndpointDict"] = [DEFAULT_METRICS_ENDPOINT],      
```

_Originally posted by @PietroPasotti in https://github.com/canonical/grafana-agent-k8s-operator/pull/95#discussion_r1127742556_

PietroPasotti:
Adds typing annotation for `COSAgentProvider`'s `metrics_endpoints` init arg.

Fixes #106

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -29,7 +29,7 @@ The constructor of `COSAgentProvider` has only one required and eight optional p
         self,
         charm: CharmType,
         relation_name: str = DEFAULT_RELATION_NAME,
-        metrics_endpoints: Optional[List[dict]] = None,
+        metrics_endpoints: Optional[List[_MetricsEndpointDict]] = None,
         metrics_rules_dir: str = "./src/prometheus_alert_rules",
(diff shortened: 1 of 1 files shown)


simskij (approved):

Abuelodelanada (approved):

dstathis (commented):
Can you explain a bit why this is useful? It seems to me like it adds code and doesn't do anything?

rbarry82:
It kind-of doesn't, but OF uses `pyright`, not `mypy`, and we're trying to slowly converge a little. `pyright` doesn't like `dict`. It could be ok with `Dict[str, Any]` or something, but it may even allow us to do a little enforcement someday. Mostly this is for pyright and a little sanity.

PR #152, opened 2023-03-23
Cleanup

sed-i:
## Issue
Need to cleanup.

## Solution
- Remove stray built-in files that were copied to root folder by tests.
- Decrement LIBPATCH (`Library charms.grafana_agent.v0.cos_agent LIBPATCH number was incorrectly incremented, Charmhub has the same content in version 0.2.`)
- Lint


rbarry82:
Don't we actually want to _keep_ those dashboards, or is there yet another node_exporter dashboard in the charm?

sed-i:
The ones I removed are just copied from `src/`. The ones in `src/` are still there.
Or did you mean something else?

rbarry82:
Nope, just missed the fact that the path _didn't_ have `src/` in front of it

rbarry82 (approved):

Issue #131, opened 2023-03-15
Provide sensible values for the `positions_dir` config


simskij:
### Enhancement Proposal

For machines, it should be in `$SNAP_DATA`.
For k8s, it doesn't really matter where it is, as long as it's not in `/tmp`.

sed-i:
Depends on snap changes so it reads env vars.

lucabello:
This has been done by the linked PR

PR #155, opened 2023-03-24
use better locations for the positions files

dstathis:
## Issue
#131

## Solution
Each charm class defines a `positions_dir` method to be called by the parent.

## Testing Instructions
Deploy on k8s and machine models and check that the positions files are in the correct place.

## Release Notes
Fixed a bug where grafana-agent-k8s was not storing position files correctly.

### DIFF
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -216,6 +216,10 @@ class GrafanaAgentCharm(CharmBase):
         """Return a list of dashboards."""
         raise NotImplementedError("Please override the dashboards method")
 
+    def positions_dir(self) -> str:
+        """Return the correct positions directory."""
+        raise NotImplementedError("Please override the positions_dir method")
+
     # End: Abstract Methods
 
     def _update_metrics_alerts(self):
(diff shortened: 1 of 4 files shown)


src/grafana_agent.py
```diff
@@ -216,6 +216,10 @@ def dashboards(self) -> list:
         """Return a list of dashboards."""
         raise NotImplementedError("Please override the dashboards method")
 
+    def positions_dir(self) -> str:
+        """Return the correct positions directory."""
```
sed-i, line 220: **nitpick:** `correct` is redundant.
```suggestion
        """Return the positions directory."""
```

src/k8s_charm.py
```diff
@@ -146,6 +146,10 @@ def restart(self) -> None:
         """Restart grafana agent."""
         self._container.restart("agent")
 
+    def positions_dir(self) -> str:
+        """Return the correct positions directory."""
```
sed-i, line 150: **nitpick:** `correct` is redundant.
```suggestion
        """Return the positions directory."""
```

src/machine_charm.py
```diff
@@ -464,6 +464,10 @@ def _connect_logging_snap_endpoints(self):
                 logger.error(f"error connecting plug {plug} to grafana-agent:logs")
                 logger.error(e.message)
 
+    def positions_dir(self) -> str:
+        """Return the correct positions directory."""
```
sed-i, line 468: **nitpick:** `correct` is redundant.
```suggestion
        """Return the positions directory."""
```

sed-i (approved):

PR #159, opened 2023-03-28
Fix fstab parser

PietroPasotti:
@MiaAltieri found a :bug:

Issue:
When fstab looks like:
```
        /var/snap/charmed-kafka/common/log /snap/grafana-agent/7/shared-logs/log none bind,ro 0 0\n
        /var/snap/charmed-kafka/common/aei /snap/grafana-agent/7/shared-logs/aei none bind,ro 0 0\n
```

I.e. multiple entries end up in `SnapFstab.entries`, then the codepath at `machine_charm.py:95` kicks in.
As a result, a TypeError comes up because `ordered_entries` attempted to use `_SnapFstabEntry` (unhashable) as dict key, which is a no-can-do.

That path was untested.

Solution:
Refactor to avoid using `_SnapFstabEntry` instances as dict keys, and add tests for that path.

:palm_tree:


src/machine_charm.py
```diff
@@ -302,14 +304,14 @@ def _additional_integrations(self) -> Dict[str, Any]:
...
-                ]
-                + self._principal_relabeling_config,
+                                       # Align the "job" name with those of prometheus_scrape
+                                       {
+                                           "target_label": "job",
+                                           "regex": "(.*)",
+                                           "replacement": node_exporter_job_name,
+                                       },
+                                   ]
+                                   + self._principal_relabeling_config,
```
simskij, line 314: This looks like an odd indentation 😅

simskij (changes requested):

src/machine_charm.py
PietroPasotti: oh what happened there

simskij (approved):

PR #171, opened 2023-04-17
Split into k8s, machine dirs

sed-i:
## Issue
1. Our CI currently does not support pseudo-hybrid charms - we would need to `render-k8s`/`-machine` in a centralized way. As a result, the `release` workflow is currently broken, and `unit-machine` and `integration-machine` are not running on PRs.
2. README.md currently reflect the k8s charm and would have to be rewritten so it addresses both.
3. Must remember to `render` before testing / manually publishing.

## Solution
- Split into separate charm libs.
- Use the `charm-path` arg in CI.

NOTES:
- This PR **depends on charmcraft working with symlinks**: https://github.com/canonical/charmcraft/issues/1079. If the charmcraft team rejects the proposal, we could duplicate the built-in files and use a hardlink for `grafana-charm.py` (can't use hardlinks for dirs).
- Itests will currently fail because charmcraft does not create the symlink content under `src/`.

## Context
NTA.

## Testing Instructions
Run tox in both subdirs.

## Release Notes
Split into k8s, machine dirs.


sed-i:
This is impl'd in:
- https://github.com/canonical/grafana-agent-k8s-operator/pull/254
- https://github.com/canonical/grafana-agent-operator/pull/2
Closing.

PR #179, opened 2023-04-27
use ruff instead of flake8

lucabello:
The **flake8** plugins we use are re-implemented in **ruff**, which is why they have been removed from the dependencies.

**isort** is enabled (and used when formatting) through the `I001` rule.

The "exclude list" has been updated to omit folders that **ruff** excludes by default.

Everything else is the same, the only difference being one new rule being ignored (which wasn't checked before anyway):
* `RET504: Unnecessary variable assignment before return statement`, which we do everywhere as it increases readability.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,26 +13,25 @@ show_missing = true
 line-length = 99
 target-version = ["py38"]
 
-[tool.isort]
-profile = "black"
-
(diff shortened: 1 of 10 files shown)


lucabello:
As usual, Check libraries is failing only because there is no `metadata.yaml` :)

PR #175, opened 2023-04-20
fix log directory traversal for machine charm

lucabello:
## Issue
Closes #163.

## Solution
The `__path__` of the log files can be a glob; however, the globbing library will always include both files and directories. `**` alone will include the files (which is the desired behavior), but won't remove the `file is directory` errors.  
Thus, a [drop pipeline](https://grafana.com/docs/loki/latest/clients/promtail/stages/drop/) is added to drop those messages specifically from the data flow.

## Testing Instructions
Deploy according to the issue description and verify those messages don't appear anymore.

## Release Notes
Correctly traverse log directories for the machine charm.

Issue #163, opened 2023-03-31
Machine charm does not currently traverse log directories


simskij:
### Bug Description

If a snap slot provides one or multiple nested folders for logs, we currently do not properly recurse these directories to find files. See the attached screenshot

![image](https://user-images.githubusercontent.com/1596025/229156938-4fac98fa-020f-4ae4-b113-401ec813a80a.png)

### To Reproduce

1. `juju deploy zookeeper`, built from source
2. `juju deploy grafana-agent`, built from source
3. `juju relate grafana-agent zookeeper`
4. relate via CMRs to Grafana

### Environment

- 

### Relevant log output

```shell
2023-03-31 16:56:57	
Mar 31 14:56:56 juju-8c976f-6 grafana-agent.grafana-agent[10431]: ts=2023-03-31T14:56:56.954835365Z caller=filetarget.go:317 level=info component=logs logs_config=log_file_scraper msg="failed to tail file" error="file is a directory" filename=/snap/grafana-agent/11/shared-logs/zookeeper/version-2
2023-03-31 16:56:56	
ts=2023-03-31T14:56:56.954835365Z caller=filetarget.go:317 level=info component=logs logs_config=log_file_scraper msg="failed to tail file" error="file is a directory" filename=/snap/grafana-agent/11/shared-logs/zookeeper/version-2
```

lucabello:
Turns out that the `__path__` configuration accepts a glob which is parsed using this [bmatcuk/doublestar](https://github.com/bmatcuk/doublestar) library.

The issue with this is that it doesn't have a way to match files only, so a couple of possible behaviors are:
1. `__path__: .../zookeeper/**` gets all files AND directories, which doesn't get rid of the "file is a directory" error, but it will at least also scrape all the files inside it
2. `__path__: .../zookeeper/**/*.log` get all files with a certain extension, getting rid of the error but limiting ourselves to some file extension for logs in that folder tree

A different solution would be to create a job for each file and pointing the `__path__` to each; having one for each folder would be equivalent to **1** if a folder contains both log files and folders.

simskij:
> Turns out that the `__path__` configuration accepts a glob which is parsed using this [bmatcuk/doublestar](https://github.com/bmatcuk/doublestar) library.
> 
> The issue with this is that it doesn't have a way to match files only, so a couple of possible behaviors are:
> 
[...]

PR #181, opened 2023-04-28
fix crashes from log glob change

lucabello:
## Issue
After tackling #163 with #175, the Grafana Agent snap started crashing. Turns out that the specified pipeline stage is causing the issue:
```
Apr 28 10:12:13 juju-5a1d83-0 grafana-agent.grafana-agent[15458]: ts=2023-04-28T10:12:13.077284721Z caller=main.go:64 level=error msg="error creating the agent server entrypoint" err="unable to apply config for log_file_scraper: unable to create logs instance: failed to make file target manager: invalid drop stage config: 1 error(s) decoding:\n\n* 'source' expected type 'string', got unconvertible type '[]interface {}', value: '[level msg]'"
```

Regardless of the pipeline stage [documentation](https://grafana.com/docs/loki/latest/clients/promtail/stages/drop/) specifying it does accept a list of values, the agent doesn't take it.

## Solution
Remove the `source` selector; the `expression` will match against the whole log line, achieving the same effect. The change has also been applied to the other scrape jobs which would otherwise report the same message.

### DIFF
--- a/src/machine_charm.py
+++ b/src/machine_charm.py
@@ -323,6 +323,13 @@ class GrafanaAgentMachineCharm(GrafanaAgentCharm):
                 "scrape_configs": [
                     {
                         "job_name": "varlog",
+                        "pipeline_stages": [
+                            {
+                                "drop": {
+                                    "expression": ".*file is a directory.*",
+                                },
+                            },
+                        ],
                         "static_configs": [
                             {
                                 "targets": ["localhost"],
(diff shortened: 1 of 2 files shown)


rbarry82 (approved):

sed-i:
For future reference:

> It's because we try to get everything in /var/log, but folders are considered as files and thus trying to get them causes that warning message, which ends up flooding the data -- @lucabello

PR #191, opened 2023-05-12
Add Prometheus rules for pressure stall information.

chanchiwai-ray:
## Issue
Moving OS-level NRPE checks (CPU related) from [charm-nrpe](https://git.launchpad.net/charm-nrpe/tree/files/plugins).

## Solution
Adding reasonable default Prometheus alert rule in Granada agent charm.

## Context
Create alert rules based on [pressure stall information (PSI)](https://www.kernel.org/doc/html/latest/accounting/psi.html)

## Testing Instructions
Tested with `promtool`
```yaml
rule_files:
  - pressure.rules

evaluation_interval: 1m

tests:

  - interval: 1m
    input_series:
      - series: 'node_pressure_cpu_waiting_seconds_total{instance="test-instance-1", job="node"}'
        values: '0+20x15' # rate = 0.33
      - series: 'node_pressure_io_waiting_seconds_total{instance="test-instance-1", job="node"}'
        values: '0+20x15' # rate = 0.33
      - series: 'node_pressure_memory_waiting_seconds_total{instance="test-instance-1", job="node"}'
        values: '0+20x15' # rate = 0.33

      - series: 'node_pressure_cpu_waiting_seconds_total{instance="test-instance-2", job="node"}'
        values: '0+60x15' # rate = 1.0 !
      - series: 'node_pressure_io_waiting_seconds_total{instance="test-instance-2", job="node"}'
        values: '0+20x15' # rate = 0.33
      - series: 'node_pressure_memory_waiting_seconds_total{instance="test-instance-2", job="node"}'
        values: '0+20x15' # rate = 0.33

      - series: 'node_pressure_cpu_waiting_seconds_total{instance="test-instance-3", job="node"}'
        values: '0+20x15' # rate = 0.33
      - series: 'node_pressure_io_waiting_seconds_total{instance="test-instance-3", job="node"}'
        values: '0+40x15' # rate = 0.67 !
      - series: 'node_pressure_memory_waiting_seconds_total{instance="test-instance-3", job="node"}'
        values: '0+20x15' # rate = 0.33

      - series: 'node_pressure_cpu_waiting_seconds_total{instance="test-instance-4", job="node"}'
        values: '0+20x15' # rate = 0.33
      - series: 'node_pressure_io_waiting_seconds_total{instance="test-instance-4", job="node"}'
        values: '0+20x15' # rate = 0.33
      - series: 'node_pressure_memory_waiting_seconds_total{instance="test-instance-4", job="node"}'
        values: '0+40x15' # rate = 0.67 !

    alert_rule_test:

      - eval_time: 6m
        alertname: HostHighCpuWaitingTime
        exp_alerts:
          - exp_labels:
              severity: info
              instance: test-instance-2
              job: node
            exp_annotations:
              summary: Host processes spent too much time waiting for CPU resources (instance test-instance-2)
              description: The instantanoues time that the processes spent on waiting for CPU resource is too high. This might indicates that the server is under high CPU pressure.\n  VALUE = 1.00\n  LABELS = map[instance:test-instance-2 job:node]

      - eval_time: 6m
        alertname: HostHighIOWaitingTime
        exp_alerts:
          - exp_labels:
              severity: info
              instance: test-instance-3
              job: node
            exp_annotations:
              summary: Host processes spent too much time waiting due to I/O congestion (instance test-instance-3)
              description: The instantanoues time that the processes spent on waiting for I/O is too high. This might indicates that the server is under high I/O pressure.\n  VALUE = 0.67\n  LABELS = map[instance:test-instance-3 job:node]

      - eval_time: 6m
        alertname: HostHighMemoryWaitingTime
        exp_alerts:
          - exp_labels:
              severity: info
              instance: test-instance-4
              job: node
            exp_annotations:
              summary: Host processes spent too much time waiting for memory (instance test-instance-4)
              description: The instantanoues time that the processes spent on waiting for memory is too high. This might indicates that the server is under high memory pressure.\n  VALUE = 0.67\n  LABELS = map[instance:test-instance-4 job:node]
```

Result

```bash
Unit Testing:  test_pressure.yaml
  SUCCESS
```

## Release Notes
- Add Prometheus alert group: `HostPressure`
    - alert rule for `HostHighCpuWaitingTime`
    - alert rule for `HostHighIOWaitingTime`
    - alert rule for `HostHighMemoryWaitingTime`

### DIFF
--- a/src/prometheus_alert_rules/pressure.rules
+++ b/src/prometheus_alert_rules/pressure.rules
@@ -7,7 +7,7 @@ groups:
     expr: irate(node_pressure_cpu_waiting_seconds_total[5m]) > 0.9
     for: 5m
     labels:
-      severity: info
+      severity: warning
     annotations:
       summary: Host processes spent too much time waiting for CPU resources (instance {{ $labels.instance }})
       description: The instantanoues time that the processes spent on waiting for CPU resource is too high. This might indicates that the server is under high CPU pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
@@ -17,7 +17,7 @@ groups:
     expr: irate(node_pressure_io_waiting_seconds_total[5m]) > 0.5
     for: 5m
     labels:
-      severity: info
+      severity: warning
     annotations:
       summary: Host processes spent too much time waiting due to I/O congestion (instance {{ $labels.instance }})
       description: The instantanoues time that the processes spent on waiting for I/O is too high. This might indicates that the server is under high I/O pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
@@ -27,7 +27,7 @@ groups:
     expr: irate(node_pressure_memory_waiting_seconds_total[5m]) > 0.5
     for: 5m
     labels:
-      severity: info
+      severity: warning
     annotations:
       summary: Host processes spent too much time waiting for memory (instance {{ $labels.instance }})
       description: The instantanoues time that the processes spent on waiting for memory is too high. This might indicates that the server is under high memory pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
--- a/src/prometheus_alert_rules/pressure.rules
+++ b/src/prometheus_alert_rules/pressure.rules
@@ -1,0 +1,33 @@
+groups:
+- name: HostPressure
+  rules:
+
+  # Alert for host cpu pressure - high instantanoues cpu waiting time
+  - alert: HostHighCpuWaitingTime
+    expr: irate(node_pressure_cpu_waiting_seconds_total[5m]) > 0.9
+    for: 5m
+    labels:
+      severity: info
+    annotations:
+      summary: Host processes spent too much time waiting for CPU resources (instance {{ $labels.instance }})
+      description: The instantanoues time that the processes spent on waiting for CPU resource is too high. This might indicates that the server is under high CPU pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
+
+  # Alert for host io pressure - high instantanoues io waiting time
+  - alert: HostHighIOWaitingTime
+    expr: irate(node_pressure_io_waiting_seconds_total[5m]) > 0.5
+    for: 5m
+    labels:
+      severity: info
+    annotations:
+      summary: Host processes spent too much time waiting due to I/O congestion (instance {{ $labels.instance }})
+      description: The instantanoues time that the processes spent on waiting for I/O is too high. This might indicates that the server is under high I/O pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
+
+  # Alert for host mem pressure - high instantanoues mem waiting time
+  - alert: HostHighMemoryWaitingTime
+    expr: irate(node_pressure_memory_waiting_seconds_total[5m]) > 0.5
+    for: 5m
+    labels:
+      severity: info
+    annotations:
+      summary: Host processes spent too much time waiting for memory (instance {{ $labels.instance }})
+      description: The instantanoues time that the processes spent on waiting for memory is too high. This might indicates that the server is under high memory pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}


src/prometheus_alert_rules/cpu.rules
```diff
@@ -0,0 +1,27 @@
+groups:
+- name: HostCPU
+  rules:
+
+  # Compute number of cores
+  - record: node_cpu_cores_total
+    expr: count(node_cpu_seconds_total{mode="idle"}) without (cpu,mode)
```
sed-i, line 7: This is probably the first time a PR introduces recording rules to our set of built-in rules.
This is a good opportunity to discuss best practices.
1. [Official docs](https://prometheus.io/docs/practices/rules/#recording-rules) recommend naming in the general form `level:metric:operations`. I think in this case it would be `instance:node_cpu_seconds:count`.
2. Recording rules are meant for massively recurring expensive operations. Seems like here it is used as an alias to simplify other _alert_ expressions. Not sure where the best tradeoff lies.

@rbarry82 @simskij

src/prometheus_alert_rules/cpu.rules
```diff
@@ -0,0 +1,27 @@
...
+
+  # Alert for cpu load average per core on 5 minute interval: warning
+  - alert: HostCpuHighLoadAvg
+    expr: node_load1 / node_cpu_cores_total >= 8 * node_procs_running OR node_load5 / node_cpu_cores_total >= 4 * node_procs_running OR node_load15 / node_cpu_cores_total >= 2 * node_procs_running
+    for: 3m
+    labels:
+      severity: critical
```
sed-i, line 14: 1. Perhaps there is a more direct (or indirect?) way to calculate high load, i.e. without needing to separately calculate number of cores?
2. Threshold are application specific, so should the severity here be `info`?

chanchiwai-ray:
Thanks for the comments. We are thinking about using [Pressure Stall information (PSI)](https://www.kernel.org/doc/html/latest/accounting/psi.html) instead of `loadavg` which is a bit old. Presumably, we will be using `irate(node_pressure_cpu_waiting_seconds_total[1m])` for the `expr`. I will revisit this PR later once we have a reference threshold for the alert. Or If you have some idea in mind, please feel free to suggest

src/prometheus_alert_rules/cpu.rules
simskij: Calculating it through a recording rule also means another time series will be created. In this particular case, that time series will be of negligible size, so it does not really matter much. 

Will you want to use this recorded metric in your dashboards? If yes, it might make sense. If not, might be just as good to inline it.

simskij: Since all of these expressions have been deleted, I'll go ahead and resolve the conversations.

src/prometheus_alert_rules/pressure.rules
```diff
@@ -0,0 +1,33 @@
+groups:
+- name: HostPressure
+  rules:
+
+  # Alert for host cpu pressure - high instantanoues cpu waiting time
+  - alert: HostHighCpuWaitingTime
+    expr: irate(node_pressure_cpu_waiting_seconds_total[5m]) > 0.9
+    for: 5m
+    labels:
+      severity: info
```
simskij, line 10: Why info? What does this imply for you specifically? I would suggest going with warning.

src/prometheus_alert_rules/pressure.rules
```diff
@@ -0,0 +1,33 @@
...
+    annotations:
+      summary: Host processes spent too much time waiting for CPU resources (instance {{ $labels.instance }})
+      description: The instantanoues time that the processes spent on waiting for CPU resource is too high. This might indicates that the server is under high CPU pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
+
+  # Alert for host io pressure - high instantanoues io waiting time
+  - alert: HostHighIOWaitingTime
+    expr: irate(node_pressure_io_waiting_seconds_total[5m]) > 0.5
+    for: 5m
+    labels:
+      severity: info
```
simskij, line 20: See note above about severity

src/prometheus_alert_rules/pressure.rules
```diff
@@ -0,0 +1,33 @@
...
+    annotations:
+      summary: Host processes spent too much time waiting due to I/O congestion (instance {{ $labels.instance }})
+      description: The instantanoues time that the processes spent on waiting for I/O is too high. This might indicates that the server is under high I/O pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
+
+  # Alert for host mem pressure - high instantanoues mem waiting time
+  - alert: HostHighMemoryWaitingTime
+    expr: irate(node_pressure_memory_waiting_seconds_total[5m]) > 0.5
+    for: 5m
+    labels:
+      severity: info
```
simskij, line 30: See note above about severity

src/prometheus_alert_rules/pressure.rules
```diff
@@ -0,0 +1,33 @@
...
+    annotations:
+      summary: Host processes spent too much time waiting due to I/O congestion (instance {{ $labels.instance }})
+      description: The instantanoues time that the processes spent on waiting for I/O is too high. This might indicates that the server is under high I/O pressure.\n  VALUE = {{ $value | printf "%.2f" }}\n  LABELS = {{ $labels }}
+
+  # Alert for host mem pressure - high instantanoues mem waiting time
+  - alert: HostHighMemoryWaitingTime
+    expr: irate(node_pressure_memory_waiting_seconds_total[5m]) > 0.5
+    for: 5m
```
simskij, line 28: Setting it to `5m` here means that not only does the rate over the last five minutes have to exceed 0.5, but this also needs to remain true for the coming five minutes before the alert actually fires. Is this intentional?

simskij (changes requested):

src/prometheus_alert_rules/pressure.rules
chanchiwai-ray: The idea is to provide user some kind alert (anything other than critical) about the `cpu` and `mem` usage, and the user should spend some time to investigate what is happening when the alert is fired. So yes, I agree `warning` should be more appropriate to prompt the user to do something.

src/prometheus_alert_rules/pressure.rules
chanchiwai-ray: Yes, those alert rules were set `for: 5m` because the expressions are quite easy to reach (tested with `stress`). And although they are reached, the host system can still be fairly usable, so setting `5m` can reduce many false alarms.

PR #201, opened 2023-05-30
Set BlockedStatus when one mandatory relation is removed

### LINKED ISSUE #164: (machine) Status isn't updated correctly on relation broken
If we have all the mandatory relations in place, and then remove one of them, the status will still show "Active".

This is because on relation-broken, the relation that is going away is [still present](https://github.com/canonical/operator/issues/888) in `self.model.relations`:

https://github.com/canonical/grafana-agent-k8s-operator/blob/db550c6ead28954c8863652cb465dc85784e0f6b/src/grafana_agent.py#L325-L325

Abuelodelanada:
## Issue

This PR fix #164.

The solution is a workaround until [this issue in ops](https://github.com/canonical/operator/issues/888) is fixed

### DIFF
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -450,7 +450,7 @@ class GrafanaAgentCharm(CharmBase):
         self._grafana_dashboards_provider._reinitialize_dashboard_data(
             inject_dropdowns=False
         )  # noqa
-        self._update_status(_event)
+        self._update_status()
 
     def _enrich_endpoints(self) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
         """Add TLS information to Prometheus and Loki endpoints."""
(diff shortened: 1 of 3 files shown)


src/grafana_agent.py
```diff
@@ -387,6 +395,23 @@ def _update_status(self, *_):
...
+        for outgoing in outgoings:
+            relation = self.model.relations.get(outgoing, [])
+
+            try:
+                units = relation[0].units
+            except IndexError:
+                units = None
+
+            if relation == []:
+                continue
```
lucabello, line 408: **nitpick**: these two blocks could be swapped because you already have the relation object, so you could enter the `continue` without checking for the unit

```suggestion
            if relation == []:
                continue

            try:
                units = relation[0].units
            except IndexError:
                units = None
```

lucabello (approved):

PR #197, opened 2023-05-19
Add other useful alert rule based on NRPE

rgildein:
- add alert rule to monitor failed systemd scope units
- add alert rule to monitor number of logged users

These alert rules required to enabled following collectors:
```yaml
    enable_collectors:
      - logind
      - systemd
```

## Context
Moving rest NRPE checks from [charm-nrpe](https://git.launchpad.net/charm-nrpe/tree/files/plugins).

## Testing Instructions
Tested with
```yaml
rule_files:
  - useful.rules

evaluation_interval: 1m

tests:
  # systemd scopes
  - interval: 1m
    input_series:
      # represent all *.scope units in failed state
      - series: node_systemd_unit_state{instance="test-model_1234_test-app_test-app/0", name="failed-units.scope", state="failed"}
        values: '0x5 20 30 40 50 60'
      # represent all *.scope units in active state
      - series: node_systemd_unit_state{instance="test-model_1234_test-app_test-app/0", name="active-units.scope", state="active"}
        values: '100x5 80 70 60 50 40'
      # represent all !*.scope units
      - series: node_systemd_unit_state{instance="test-model_1234_test-app_test-app/0", name="all-other-units.socket", state="failed"}
        values: '200x10'
    alert_rule_test:
      - eval_time: 5m
        alertname: HostSystemdFailedScopes
        exp_alerts: []  # no alert
      - eval_time: 6m
        alertname: HostSystemdFailedScopes
        exp_alerts: []  # no alert
      - eval_time: 7m
        alertname: HostSystemdFailedScopes
        exp_alerts: []  # no alert
      - eval_time: 8m
        alertname: HostSystemdFailedScopes
        exp_alerts: []  # no alert
      - eval_time: 9m
        alertname: HostSystemdFailedScopes
        exp_alerts:
          - exp_labels:
              severity: critical
              instance: test-model_1234_test-app_test-app/0
            exp_annotations:
              summary:  Host has 50 systemd scopes in failed state (instance test-model_1234_test-app_test-app/0)
              description: >-
                Host has 50 systemd scopes in failed state.
                  VALUE = 50
                  LABELS = map[instance:test-model_1234_test-app_test-app/0]
      - eval_time: 10m
        alertname: HostSystemdFailedScopes
        exp_alerts:
          - exp_labels:
              severity: critical
              instance: test-model_1234_test-app_test-app/0
            exp_annotations:
              summary:  Host has 60 systemd scopes in failed state (instance test-model_1234_test-app_test-app/0)
[...]

PR #202, opened 2023-06-01
enable some additional useful collectors

simskij:
## Issue

https://github.com/canonical/grafana-agent-k8s-operator/pull/197 won't work without the optional `logind` and `systemd` collectors enabled.

## Solution

Enabling the two collectors above, as well as a couple of additional ones that could prove useful.

## Testing Instructions

Deploy and make sure the `node_systemd_unit_state` metric is properly populated.

## Release Notes

Enable additional optional collectors.


rgildein (commented):
Thanks for enabling these collectors, I checked that these are all we needed in recent alert rules. :+1:

PR #215, opened 2023-07-04
Pin pydantic's major version

### LINKED ISSUE #214: pydantic 2.0 dependency
### Bug Description

The pydantic just [released 2.0](https://github.com/pydantic/pydantic/releases/tag/v2.0).
So the charm build and and the `PY_DEPS` in library is broken.
Also maybe some fixs are requried for pydantic related code.

### To Reproduce

```
tox -e render-machine
charmcraft pack
```

### Environment

ubuntu 20.04

### Relevant log output

```shell
...
```

### Temporary workaround

Pin pydantic version < 2.0 in `PY_DEPS`

```python
# PYDEPS = ["cosl", "pydantic"]
PYDEPS = ["cosl", "pydantic < 2.0"]
```

### Additional context

_No response_

dragomirp:
## Issue
Closes #214 

When packing charmcraft will install and update dependencies of charm libraries overriding pins set in the charm's `requirements.txt`. Since there's no version restriction set in `cos-agent` PYDEPS for pydantic, charmcraft will always update and pull the latest version, which will break charms using it. 

## Solution
Pin pydantic in the cos-agent charm library to less than 2

## Context
Charmcraft issue:
https://github.com/canonical/charmcraft/issues/1135

And PR:
https://github.com/canonical/charmcraft/pull/1136

## Testing Instructions
N/A

## Release Notes
Pin pydantic's major version in cos-agent PYDEPS

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -189,9 +189,9 @@ if TYPE_CHECKING:
 
 LIBID = "dc15fa84cef84ce58155fb84f6c6213a"
 LIBAPI = 0
-LIBPATCH = 4
+LIBPATCH = 5
 
-PYDEPS = ["cosl", "pydantic"]
+PYDEPS = ["cosl", "pydantic<2"]
 
 DEFAULT_RELATION_NAME = "cos-agent"
 DEFAULT_PEER_RELATION_NAME = "peers"
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -352,7 +352,7 @@ class COSAgentProvider(Object):
                     )
                     relation.data[self._charm.unit][data.KEY] = data.json()
                 except (
-                    pydantic.error_wrappers.ValidationError,
+                    pydantic.ValidationError,
                     json.decoder.JSONDecodeError,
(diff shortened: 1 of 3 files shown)


sed-i (approved):

sed-i:
Thanks @dragomirp!
Not part of this PR, but would you like to attempt to fix the static failures?

dragomirp:
> Thanks @dragomirp! Not part of this PR, but would you like to attempt to fix the static failures?

Hi, I'll give it a try.

sed-i:
Hey @dragomirp,
Itests fail because
`juju.errors.JujuAPIError: application "agent" not found`
because
`Failed to build charm for bases index '0'.`

I'm EOD. If you get to it before me, try to pack locally?

dragomirp:
> Pinning the major version of the lib could cause issues for dependents—e.g. say a breaking change is made to pydantic that doesn't affect the charm lib. If a charm needs to use that newer version of pydantic, it will be unable to use the charm lib even though the charm lib is compatible

If the library cannot work with the newer version, as is the case here, a charm cannot really use the newer version either way. More so, I personally was unable to compile pydantic 2 on focal, so I'm not sure how easily we can make that upgrade even beyond the issue at hand. FWIW, I agree it's better to make things work with pydantic 2 and for version control to stay with the charm devs, but I'm not sure that's feasible short term.

PR #219, opened 2023-07-12
Do not override static config labels

neoaggelos:
## Issue

The grafana-agent charm will override any existing labels that may be configured in the static_configs

## Solution

Use `static_config.setdefault("labels", {})`, and then update the dict so that:

- If any `labels` are set in the static config, they are preserved and the juju topology labels are added
- If no `labels` are set in the config, this is equivalent to the previous behavior

## Context

Needed for https://github.com/canonical/charm-microk8s/pull/88 (along with a few more PRs I'm making separately)

## Testing Instructions

Supply a static config with existing labels, expect that they are not removed from the resulting scrape jobs.

## Release Notes

Respect any labels defined in the metrics endpoints.

### DIFF
--- a/src/machine_charm.py
+++ b/src/machine_charm.py
@@ -255,7 +255,11 @@ class GrafanaAgentMachineCharm(GrafanaAgentCharm):
         for job in jobs:
             static_configs = job.get("static_configs", [])
             for static_config in static_configs:
-                static_config["labels"] = self._principal_labels
+                static_config["labels"] = {
+                    # Be sure to keep labels from static_config
+                    **static_config.get("labels", {}),
+                    **self._principal_labels,
+                }
         return jobs
 
     def logs_rules(self) -> Dict[str, Any]:


src/machine_charm.py
```diff
@@ -255,7 +255,8 @@ def metrics_jobs(self) -> list:
         for job in jobs:
             static_configs = job.get("static_configs", [])
             for static_config in static_configs:
-                static_config["labels"] = self._principal_labels
+                static_config.setdefault("labels", {})
+                static_config["labels"].update(**self._principal_labels)
```
sed-i, line 259: TIL about `setdefault`, but do you think this is more straightforward?
```suggestion
                static_config["labels"] = {
                    # Be sure to keep labels from static_config
                    **static_config.get("labels", {}),
                    **self._principal_labels,
                }
```

neoaggelos: I had the foggiest memory that `{**dict1, **dict2}` breaks if there is a common set of keys, but apparently this does not seem to be the case:

```
>>> a = {"a": "b"}
>>> b = {"a": "c"}
>>> {**a, **b}
{'a': 'c'}
```

Sure, no problem to do it like this :+1:

sed-i (approved):

PR #220, opened 2023-07-12
Allow custom scrape jobs in COSAgentProvider

neoaggelos:
## Issue

Scrape jobs may require more fields, e.g. `authorization` or custom labels for specific targets

## Solution

If the metrics endpoint passed to `COSAgentProvider` has a `scrape_job` field, then assume that this is a custom job, also added docs (happy to adjust with any recommendation)

## Context

Needed for https://github.com/canonical/charm-microk8s/pull/88, also see to #219 

## Release Notes

Allow custom scrape jobs in COSAgentProvider

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -22,20 +22,21 @@ this charm library.
 Using the `COSAgentProvider` object only requires instantiating it,
 typically in the `__init__` method of your charm (the one which sends telemetry).
 
-The constructor of `COSAgentProvider` has only one required and eight optional parameters:
+The constructor of `COSAgentProvider` has only one required and nine optional parameters:
 
 ```python
     def __init__(
         self,
         charm: CharmType,
         relation_name: str = DEFAULT_RELATION_NAME,
-        metrics_endpoints: Optional[Union[List[_MetricsEndpointDict], Callable]] = None,
+        metrics_endpoints: Optional[List[_MetricsEndpointDict]] = None,
         metrics_rules_dir: str = "./src/prometheus_alert_rules",
         logs_rules_dir: str = "./src/loki_alert_rules",
         recurse_rules_dirs: bool = False,
         log_slots: Optional[List[str]] = None,
         dashboard_dirs: Optional[List[str]] = None,
         refresh_events: Optional[List] = None,
+        scrape_configs: Optional[Union[List[Dict], Callable]] = None,
     ):
 ```
 
@@ -47,7 +48,8 @@ The constructor of `COSAgentProvider` has only one required and eight optional p
     the `cos_agent` interface, this is where you have to specify that.
 
 - `metrics_endpoints`: In this parameter you can specify the metrics endpoints that Grafana Agent
-    machine Charmed Operator will scrape.
+    machine Charmed Operator will scrape. The configs of this list will be merged with the configs
+    from `scrape_configs`.
 
 - `metrics_rules_dir`: The directory in which the Charmed Operator stores its metrics alert rules
   files.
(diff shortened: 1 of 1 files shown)


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -91,9 +91,25 @@ def __init__(self, *args):
...
+                        "authorization": {"credentials": "bearer-token"},
+                        "static_configs": [
+                            {
+                                "targets": ["localhost:9000"]},
+                                "labels": {"key": "value"},
+                            },
+                        ],
+                    },
+                }
             ],
```
sed-i, line 113: I like this.
1. I'm not a fan of "implicit dual schema". I'd rather bump LIBAPI and only have the "scrape_job" option.
2. Iiuc, we could probably omit the "scrape_job" key (it's actually "[scrape_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)"). 

```suggestion
            scrape_configs=[
                {
                        "job_name": "custom_job",
                        "metrics_path": "/metrics",
                        "authorization": {"credentials": "bearer-token"},
                        "static_configs": [
                            {
                                "targets": ["localhost:9000"]},
                                "labels": {"key": "value"},
                            },
                        ],
                },
            ],
```
Wdyt?
cc: @simskij @Abuelodelanada @dstathis @lucabello @PietroPasotti

neoaggelos: I'm 100% in on this, I understand the `metrics_endpoints` was only added for simplicity. Let me know how to proceed

sed-i: Wdyt about this [commit](https://github.com/canonical/grafana-agent-k8s-operator/pull/220/commits/da96b8f005cbb08e73234387fd7c0fb15e4a7230)?

sed-i: After merging one of your other PRs we now have a conflict with the Callable.
I'm leaving this for you to resolve as I'm not sure which solution best fits your use case.

neoaggelos: > Wdyt about this [commit](https://github.com/canonical/grafana-agent-k8s-operator/pull/220/commits/da96b8f005cbb08e73234387fd7c0fb15e4a7230)?

Looks great, let me get back to this to fix the conflicts and I'll ping you. Thanks a lot!

lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -677,15 +721,7 @@ def metrics_jobs(self) -> List[Dict]:
...
-            jobs = data.metrics_scrape_jobs
-            if jobs:
-                for job in jobs:
-                    job_config = {
-                        "job_name": job["job_name"],
-                        "metrics_path": job["path"],
-                        "static_configs": [{"targets": [f"localhost:{job['port']}"]}],
-                    }
-                    scrape_jobs.append(job_config)
+            scrape_jobs = data.metrics_scrape_jobs
```
neoaggelos, line 725: This is not backwards-compatible, does it take bumping the LIBAPI version or just handle the old format as well?

lib/charms/grafana_agent/v0/cos_agent.py
neoaggelos: Updated, but left a comment elsewhere. I like being explicit with a `scrape_configs` for this :+1:

sed-i (approved):

lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -677,15 +721,16 @@ def metrics_jobs(self) -> List[Dict]:
         """Parse the relation data contents and extract the metrics jobs."""
         scrape_jobs = []
         if data := self._principal_unit_data:
-            jobs = data.metrics_scrape_jobs
-            if jobs:
-                for job in jobs:
-                    job_config = {
+            for job in data.metrics_scrape_jobs:
+                # This is to ensure backwards compatibility with old clients
```
sed-i, line 725:
```suggestion
                # In #220, relation schema changed from a simplified dict to the standard 
                # `scrape_configs`.
                # This is to ensure backwards compatibility with Providers older than v0.5.
```

PR #222, opened 2023-07-12
Always update the cos-agent relation

neoaggelos:
## Issue

The `refresh_events` field of `COSAgentProvider` are a list of events that should cause a refresh of the `cos-agent` relation data. However, sometimes a wrong relation may be updated instead of the cos-agent relation. Consider the following example:

```yaml
# metadata.yaml

peers:
  peer:
    interface: peer

provides:
  cos-agent:
    interface: cos_agent
```

```py
# src/charm.py
self._cos = COSAgentProvider(
    relation_name="cos-agent",
    refresh_events=[self.on.peer_relation_changed, self.on.upgrade_charm]
)
```

With the above, we want to update the `cos-agent` relation when a change happens to the peer relation. However, the updated unit data is written on the `peer` relation instead, since `isinstance(event, RelationEvent)` is true and `event.relation` refers to the peer relation.

I could pinpoint the problem by adding relevant log lines in the `_on_refresh` method:

```
unit-microk8s-0: 16:53:15 INFO unit.microk8s/0.juju-log Refresh triggered by <UpgradeCharmEvent via MicroK8sCharm/on/upgrade_charm[1081]>
unit-microk8s-0: 16:53:17 INFO unit.microk8s/0.juju-log Writing into <ops.model.Relation cos-agent:11>
unit-microk8s-0: 16:53:17 INFO juju.worker.uniter.operation ran "upgrade-charm" hook (via hook dispatching script: dispatch)
unit-microk8s-0: 16:53:17 INFO juju.worker.uniter found queued "config-changed" hook
unit-microk8s-0: 16:53:20 INFO juju.worker.uniter.operation ran "config-changed" hook (via hook dispatching script: dispatch)
unit-microk8s-0: 16:53:54 INFO unit.microk8s/0.juju-log peer:0: Refresh triggered by <RelationChangedEvent via MicroK8sCharm/on/peer_relation_changed[1087]>
unit-microk8s-0: 16:53:57 INFO unit.microk8s/0.juju-log peer:0: Writing into <ops.model.Relation peer:0>
unit-microk8s-0: 16:53:58 INFO juju.worker.uniter.operation ran "peer-relation-changed" hook (via hook dispatching script: dispatch)                                       
```
Note that the right relation data is updated when the `upgrade_charm` event is fired, since it is not a relation event.

## Solution

The unit data must always be written to the `cos-agent` relation unit data. After the change of this PR, the logs are now:

```
unit-microk8s-0: 21:30:25 INFO juju.worker.uniter.operation ran "update-status" hook (via hook dispatching script: dispatch)                                    
unit-microk8s-0: 21:30:27 INFO unit.microk8s/0.juju-log peer:0: Refresh triggered by <RelationChangedEvent via MicroK8sCharm/on/peer_relation_changed[1417]>    
unit-microk8s-0: 21:30:29 INFO unit.microk8s/0.juju-log peer:0: Writing into <ops.model.Relation cos-agent:11>                                                  
unit-microk8s-0: 21:30:29 INFO juju.worker.uniter.operation ran "peer-relation-changed" hook (via hook dispatching script: dispatch)                            
unit-grafana-agent-5: 21:30:35 INFO juju.worker.uniter.operation ran "cos-agent-relation-changed" hook (via hook dispatching script: dispatch)                  
unit-grafana-agent-5: 21:33:03 INFO juju.worker.uniter.operation ran "update-status" hook (via hook dispatching script: dispatch)                               
```

We also see that the grafana-agent unit picks up the change this time.

## Context

Needed for https://github.com/canonical/charm-microk8s/pull/88

## Release Notes

Properly update cos-agent relation unit data when a relation event is used as `refresh_event`.

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -171,7 +171,7 @@ from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Optional, Set, Unio
 import pydantic
 from cosl import JujuTopology
 from cosl.rules import AlertRules
-from ops.charm import RelationChangedEvent, RelationEvent
+from ops.charm import RelationChangedEvent
 from ops.framework import EventBase, EventSource, Object, ObjectEvents
 from ops.model import Relation, Unit
 from ops.testing import CharmType
@@ -331,10 +331,7 @@ class COSAgentProvider(Object):
 
     def _on_refresh(self, event):
         """Trigger the class to update relation data."""
-        if isinstance(event, RelationEvent):
-            relations = [event.relation]
-        else:
-            relations = self._charm.model.relations[self._relation_name]
+        relations = self._charm.model.relations[self._relation_name]
 
         for relation in relations:
             # Before a principal is related to the grafana-agent subordinate, we'd get


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -331,10 +331,7 @@ def __init__(
 
     def _on_refresh(self, event):
         """Trigger the class to update relation data."""
-        if isinstance(event, RelationEvent):
-            relations = [event.relation]
-        else:
-            relations = self._charm.model.relations[self._relation_name]
+        relations = self._charm.model.relations[self._relation_name]
```
sed-i, line 334: This would work.
I think the original intention was to reduce unnecessary rel-changed events.
Lists could be unsorted and result in redundant rel-changed events.
In that case we could augment the conditional:
```python
        if isinstance(event, RelationEvent) and event.relation.name == self._relation_name:
            relations = [event.relation]
        else:
            relations = self._charm.model.relations[self._relation_name]
```
On the other hand, I do prefer holistic over deltas.
Wdyt?

sed-i (approved):

dstathis (approved):

lib/charms/grafana_agent/v0/cos_agent.py
neoaggelos: So, to me it looks like the events of the `cos-agent` relation are not relevant. The way I see it, the `refresh_events` are a list of events **other than** the `cos_agent_relation_changed` (so the second conditional would always be false).

Happy to adjust to this though, if you think that this is better. I just do not understand why a charm would be setting the `cos_agent_relation_changed` event as a refresh event (I understand that this would instead happen on the consumer side only).

neoaggelos: I don't have a strong opinion on this, for me it's enough that the `cos-agent` relation is updated :)

neoaggelos: Definitely reads a lot better to me when it says "update the `cos-agent` relations", though

PR #223, opened 2023-07-12
Support a callable for COSAgentProvider.metrics_endpoints

neoaggelos:
## Issue

It may be desired for the metrics endpoints to be generated independently of the lifecycle of the `COSAgentProvider` object. Some use cases are:

- retrieve a bearer token
- retrieve data from other relations, e.g. tokens or certificates

## Solution

Allow `metrics_endpoints` to be a callable instead of a static list. It will be executed only when the object needs to update the scrape jobs.

## Context

Needed for https://github.com/canonical/charm-microk8s/pull/88

## Release Notes

Allow a callable to dynamically generate the metrics endpoints

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -104,6 +104,31 @@ class TelemetryProviderCharm(CharmBase):
         )
 ```
 
+### Example 3 - Dynamic metrics endpoints generation:
+
+Pass a function to the `metrics_endpoints` to decouple the generation of the endpoints
+from the instantiation of the COSAgentProvider object.
+
+```python
+from charms.grafana_agent.v0.cos_agent import COSAgentProvider
+...
+
+class TelemetryProviderCharm(CharmBase):
+    def generate_metrics_endpoints(self):
+        return [
+            {"path": "/metrics", "port": 9000},
+            {"path": "/metrics", "port": 9001},
+            {"path": "/metrics", "port": 9002},
+        ]
+
+    def __init__(self, *args):
+        ...
+        self._grafana_agent = COSAgentProvider(
+            self,
+            metrics_endpoints=self.generate_metrics_endpoints,
+        )
+```
+
 ## COSAgentConsumer Library Usage
 
 This object may be used by any Charmed Operator which gathers telemetry data by
@@ -166,7 +191,7 @@ import lzma
 from collections import namedtuple
 from itertools import chain
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Optional, Set, Union
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Optional, Set, Union, Callable
 
 import pydantic
 from cosl import JujuTopology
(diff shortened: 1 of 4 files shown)


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -361,9 +388,15 @@ def _on_refresh(self, event):
     def _scrape_jobs(self) -> List[Dict]:
         """Return a prometheus_scrape-like data structure for jobs."""
         job_name_prefix = self._charm.app.name
+
+        if isinstance(self._metrics_endpoints, Callable):
```
sed-i, line 392:
```suggestion
        if callable(self._metrics_endpoints):
```

lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -302,6 +327,8 @@ def __init__(
             charm: The `CharmBase` instance that is instantiating this object.
             relation_name: The name of the relation to communicate over.
             metrics_endpoints: List of endpoints in the form [{"path": path, "port": port}, ...].
+                A callable that returns the list may be passed in case the endpoints need to be
+                generated dynamically.
```
sed-i, line 331: - Would you be able to utilize the `refresh_events` arg instead of adding the `Callable` functionality?
- The type hint for the constructor arg needs to be update accordingly. I'm on the fence for whether we should bump LIBAPI and only use a callable instead of a Union.

sed-i (commented):
I wonder if you would be able to utilize the `refresh_events` arg instead of adding the `Callable` functionality.

dstathis (approved):
I like this approach a lot :+1:

lib/charms/grafana_agent/v0/cos_agent.py
dstathis: * I don't think so. Not reasonably at least. If you are doing significant logic, you should be doing it in a hook and not in the charm's __init__() method.
* You are suggesting requiring it to be callable? I think most charms will just have a static list so we should allow that.

neoaggelos: Even in the case where a refresh event would be useful, it still seems to reverse the logic. I can imagine something like this:

```py
def __init__(self):
  self.on($some_relation_change, self.update_cached_thing)
  self._cos = COSAgentProvider(self, refresh_events=[self.on.$some_relation_change], metrics_endpoints=[{some data with self._state.cached_thing}])

def update_cached_thing(self, event):
  self._state.cached_thing = event.data[event.app]["something"]
```

This is a lot harder to grasp at first sight. Further, even like this, the `metrics_endpoints` have already been evaluated, therefore the changes to `self._state.cached_thing` are ignored.

I'd much rather prefer the callable approach, clearly demonstrating the intent of the code, while also making it much easier to test and reason about.

What did you have in mind @sed-i?

neoaggelos: This is also similar to the `lookaside_jobs` in the `MetricsEndpointProvider` library

neoaggelos:
I see that the integration tests are failing with:
```
  File "/home/runner/work/grafana-agent-k8s-operator/grafana-agent-k8s-operator/tests/integration/conftest.py", line 89, in prometheus_tester_charm
    charm = await ops_test.build_charm(charm_path)
  File "/home/runner/work/grafana-agent-k8s-operator/grafana-agent-k8s-operator/.tox/integration/lib/python3.10/site-packages/pytest_operator/plugin.py", line 1011, in build_charm
    raise RuntimeError(
RuntimeError: Failed to build charm tests/integration/prometheus-tester:
Packing the charm.
Launching environment to pack for base name='ubuntu' channel='20.04' architectures=['amd64'] (may take a while the first time but it's reusable)
charmcraft internal error: BaseCompatibilityError(brief="Incompatible base detected: Expected OS 'Ubuntu', found None.", details=None, resolution='Clean incompatible instance and retry the requested operation.')
```

I imagine this is unrelated?

lib/charms/grafana_agent/v0/cos_agent.py
sed-i: Yep. Let's keep the callable.
The other approach is events: the `cache_thing` would be expected to emit an event that would be passed as a refresh event.
Resolving.

PR #225, opened 2023-07-13
implement certificate_transfer requirer

lucabello:
## Issue
**grafana-agent** needs to be able to get and trust **traefik**'s certificate.

## Solution
~Use the [mutual_tls](https://github.com/canonical/mutual-tls-interface/blob/main/lib/charms/mutual_tls_interface/v0/mutual_tls.py) interface, which (besides what the - wrong - name suggests) helps transfer the Provider's certificate over relation data to the Requirers.~

The interface has now been renamed to [certificate-transfer](https://github.com/canonical/certificate-transfer-interface)

## Possible TODOs
- [ ] Ideally add an integration test with a Provider and test it

### DIFF
--- a/src/grafana_agent.py
+++ b/src/grafana_agent.py
@@ -176,7 +176,7 @@ class GrafanaAgentCharm(CharmBase):
             self._on_cert_transfer_changed,
         )
         self.framework.observe(
-            self.on.cert_transfer.relation_broken,  # pyright: ignore
+            self.on["cert_transfer"].relation_broken,  # pyright: ignore
             self._on_cert_transfer_changed,
         )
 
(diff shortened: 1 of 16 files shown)


PietroPasotti:
@lucabello what's the status on this work? The CI seems seriously stuck

src/grafana_agent.py
```diff
@@ -199,6 +206,12 @@ def _on_cloud_config_revoked(self, _) -> None:
         logger.info("cloud config revoked")
         self._update_config()
 
+    def _on_cert_transfer_available(self, event):
+        # TODO: trust the certificate instead of printing it
```
sed-i, line 210: If we do "delta-charming" here, then I imagine that with a unique filename such as `model-model_uuid-relation_name-relation_id` we should be able to also delete the file on relation broken.

lucabello:
This PR is dependent on and blocked by:
* https://github.com/canonical/certificate-transfer-interface/issues/5
* https://github.com/canonical/certificate-transfer-interface/issues/6

Otherwise, we can't save the certificate to file as we wouldn't know which relation the certificate comes from.

simskij (dismissed):
Missed this during the first review, but the k8s charm also needs the cert transfer relation

simskij (approved):

PR #228, opened 2023-07-14
TLS

### LINKED ISSUE #217: TLS support
### Enhancement Proposal

We need to support TLS for the http and grpc endpoints.

dstathis:
## Issue
Fixes #217

## Solution
Use cert_manager to enable TLS.

## Context
Note that this does not enable TLS for the machine charm. This is because we can't change the cli args of a snap service. This is not a problem because the machine charm does not expose any endpoints outside the machine.

## Testing Instructions
1. Deploy Charm
2. Relate to a logging client (such as zinc-k8s)
3. relate to self-signed-certs
4. check that the endpoints are using tls

## Release Notes
Grafana Agent Operator now supports TLS for it's loki_push_api endpoints.


src/grafana_agent.py
```diff
@@ -433,6 +453,15 @@ def _update_config(self) -> None:
             # Grafana-agent is not yet available so no need to update config
             return
 
+        # Write TLS files.
+        if self.cert.enabled:
```
PietroPasotti, line 457: shouldn't we delete the cert and key paths if not cert.enabled?

tests/unit/k8s/test_scrape_configuration.py
```diff
@@ -75,6 +75,8 @@
...
...ICATE-----\nMIIDRDCCAiygAwIBAgIUPf3TRo7siuI9UP63UKsRF/4sUakwDQYJKoZIhvcNAQEL\nBQAwOTELMAkGA1UEBhMCVVMxKjAoBgNVBAMMIXNlbGYtc2lnbmVkLWNlcnRpZmlj\nYXRlcy1vcGVyYXRvcjAeFw0yMzA3MTMyMzM2MzdaFw0yNDA3MTIyMzM2MzdaMEEx\nEDAOBgNVBAMMB2FnZW50LzAxLTArBgNVBC0MJGJhMjk3MDhkLTEwMTYtNGQ2OS04\nOTRjLTgwOGYxYjZkMzI5NzCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEB\nAJTck0+65BgohrJYrfQJKhkqiSZ1RD3upOHd69xzXda+5SJ+tjVU8aKTrY7BMyzD\n97mmeCZMryVrwa3JuoQR2ktf4Wq7DqRjG+YKFrN+MjlXu5ljTNvDZUheteahVUr9\nPDB67zgrhgc8TgorcESFlLSVQOQrMx3BEPN+PSFDqUme+0zQ5G9qCAY785ikAn7J\nGwd2tONYvrG1jA1rRw8SsF6CVZsgFgbHpfaj4Oq+yygwHXA1i0s7rf9DlUxrYaZC\nth2TzEPV9Iu0HNrJzWHJxDpeOvTptSZs5mBKSgCVW9nIaIuPWj8sEwKW9YVVlNEx\ngyjHLST4o340lKVqk2qXXE0CAwEAAaM8MDowOAYDVR0RBDEwL4ItYWdlbnQtMC5h\nZ2VudC1lbmRwb2ludHMubG1hLnN2Yy5jbHVzdGVyLmxvY2FsMA0GCSqGSIb3DQEB\nCwUAA4IBAQA8sD4zDFp0+cmv6acbSWiAVn/qZBcz0qpqmOqeUwd1IvcsczOwSzq5\n8Cf1TCJuqJFiZgyiAVNRXF8MxNXrYYB/LI5L52AeAwsv6Fpo+x8tqkzNwbpY+2aA\n1Q2Toc5/mCCSH9rzsznCjHnR1yUcEt/CzI9wfwdH0F5v5u7onrx0XXtnEiDTQIDV\nPQxGlI2xPdisqdLmFZgCjOpANRfAiBnQY1pLkH3rwHWHDsvb1jTy8UEpzYINu0vd\nogI1GMccHAq4MM7tRjmSuOL9XqgTEEvSJeJDSexDe0iWC3AYCat79tUV9gGfpkK/\nA+GdAqYgyk2tL0iL4WTIkg0t52pgkXPf\n-----END CERTIFICATE-----"]}]"""
```
PietroPasotti, line 78: can we find a way to have "foobar" in here, instead of this humongous realistic-looking cert? Or do we fail some validation we can't easily mock?

tests/unit/k8s/test_scrape_configuration.py
```diff
@@ -260,6 +262,14 @@ def test__cli_args(self):
         expected = "-config.file=/etc/grafana-agent.yaml"
         self.assertEqual(self.harness.charm._cli_args(), expected)
 
+    def test__cli_args_with_tls(self):
```
PietroPasotti, line 265:
```suggestion
    def test_cli_args_with_tls(self):
```

PietroPasotti (approved):
looks good overall, couple of nits

tests/unit/k8s/test_scrape_configuration.py
dstathis: I just pasted from relation data. I'll check.

dstathis:
Note: I do not think grafana-agent machine charm will be able to read the CA cert as written here.

tox.ini
```diff
@@ -149,18 +149,20 @@ deps =
...
-    pytest -v --tb native --log-cli-level=INFO -s {posargs} {[vars]tst_path}/scenario
+    pytest -vv --tb native --log-cli-level=INFO -s {posargs} {[vars]tst_path}/scenario
 
 [testenv:integration-machine]
 description = Run integration tests (machine charm)
 deps =
     aiohttp
     asyncstdlib
+    cryptography
+    jsonschema
```
sed-i, line 165: We shouldn't need these here. Leftover?

src/grafana_agent.py
```diff
@@ -626,6 +691,17 @@ def _loki_config(self) -> Dict[str, Union[Any, List[Any]]]:
...
+            for config in configs:
+                for scrape_config in config.get("scrape_configs", []):
+                    if scrape_config.get("loki_push_api"):
+                        scrape_config["loki_push_api"]["server"][
+                            "http_tls_config"
+                        ] = self.tls_config
+                        scrape_config["loki_push_api"]["server"][
+                            "grpc_tls_config"
+                        ] = self.tls_config
```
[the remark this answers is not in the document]
sed-i: If the agent is not related to CA but the scrape targets are, we would need to sprinkle `insecure: true`.

k8s_metadata.yaml
```diff
@@ -37,6 +37,9 @@ resources:
     description: OCI image for Grafana Agent
 
 requires:
+  certificates:
+    interface: tls-certificates
+    limit: 1
```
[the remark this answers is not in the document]
lucabello: The same would have signed all the scrape targets, but not necessarily the remote write targets, right? 

As in, **grafana-agent** and the clients it's interacting with will have the same CA, but the **prometheus** it's remote-writing to won't, which is why we'll have the cert_transfer relation to address that

src/grafana_agent.py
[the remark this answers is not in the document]
dstathis: No this is for the loki-push-api endpoint exposed by grafana-agent.

src/grafana_agent.py
```diff
@@ -687,3 +771,16 @@ def _agent_version(self) -> Optional[str]:
...
+
+    def _update_ca(self) -> None:
+        """Updates the CA cert on disk from cert_manager."""
+        if (not self.cert.enabled) or (not self.cert.ca):
+            try:
+                self.read_file(self._ca_path)
+            except (FileNotFoundError, PathError):
+                pass
+            else:
+                self.delete_file(self._ca_path)
```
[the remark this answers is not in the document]
lucabello: As a quicker fix, I'm leaving this in to avoid changing both the K8S and Machine charms, since I'm not 100% sure it won't have side-effects (maybe somewhere else in the code we assume that function will fail on a missing file)

tests/unit/k8s/test_scrape_configuration.py
```diff
@@ -260,6 +262,14 @@ def test__cli_args(self):
...
+        rel_id = self.harness.add_relation("certificates", "certs")
+        self.harness.add_relation_unit(rel_id, "certs/0")
+        expected = (
+            "-config.file=/etc/grafana-agent.yaml -server.http.enable-tls -server.grpc.enable-tls"
```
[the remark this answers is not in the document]
lucabello: We discussed this in a separate meeting; we agreed that we'd rather enable TLS as soon as the relation is added, and block (or whatever the correct status is) until the certificate arrives.

k8s_metadata.yaml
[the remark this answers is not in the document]
sed-i: Sounds right!

src/grafana_agent.py
```diff
@@ -433,6 +471,25 @@ def _update_config(self) -> None:
...
+            self.write_file(self._ca_path, self.cert.ca)
+        else:
+            try:
+                self.read_file(self._cert_path)
+            except (FileNotFoundError, PathError):
+                pass
+            else:
+                self.delete_file(self._cert_path)
+                self.delete_file(self._key_path)
+                self.delete_file(self._ca_path)
```
sed-i, line 491: The `try` only has `_cert_path` so these two could still technically raise.
But why do this at all? Why not `recursive=True` and drop the `read_file`?

tests/unit/k8s/test_scrape_configuration.py
dstathis: Fixed.

src/k8s_charm.py
```diff
@@ -157,14 +160,36 @@ def write_file(self, path: Union[str, pathlib.Path], text: str) -> None:
...
+
+    def stop(self) -> None:
+        """Stop grafana agent."""
+        self._container.stop("agent")
+
     def restart(self) -> None:
         """Restart grafana agent."""
+        self._container.add_layer(self._name, self._layer(), combine=True)
+        self._container.autostart()
         self._container.restart("agent")
```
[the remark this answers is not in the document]
dstathis: There is a difference. This method will restart the service even if there is no change. Replan will not.

src/grafana_agent.py
dstathis: there is no `recursive=True` for `delete_file` but I addressed it by checking for each file individually here.

lucabello (approved):

PR #232, opened 2023-08-03
Multiple subordinate relations

### LINKED ISSUE #210: [Machine Charm] Undeterministic when there is another subordinate charm.
### Enhancement Proposal

Currently, the charm expects there to exactly one principal relation. However, in the case of co-located subordinates, this is no longer true. To allow for this, the charm needs to properly handle and merge configuration provided both by the actual principal and any co-deployed subordinates.

The deployment scenario illustrated as a diagram:
20230615_15h01m14s_grim

dstathis:
## Issue
Fixes #210

## Solution
Added a subordinate field to the cos-agent relation interface so we could allow for more than one related app.

## Testing Instructions
Relate to hardware-observer and zookeeper at the same time. Observe that everything works as expected.

## Release Notes
Support for running alongside other subordinates.


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -493,6 +495,12 @@ class COSAgentRequirerEvents(ObjectEvents):
     validation_error = EventSource(COSAgentValidationError)
 
 
+class MultiplePrincipalsError(Exception):
+    """Custom exceptionfor when there are multiple principal applications."""
```
Abuelodelanada, line 499:
```suggestion
    """Custom exception for when there are multiple principal applications."""
```

Abuelodelanada (approved):

Issue #190, opened 2023-05-12
Alert rules are injected with incorrect label filters


chanchiwai-ray:
### Bug Description

Grafana agent charm has default alert rules for prometheus. However, when the charm is deployed and relate to COS, the alert rules are injected with incorrect JujuTopology (`juju_*`) labels, and that matches to no timeseries. 

Take for example, the `HostOomKillDetected` alert rule. In Prometheus web UI, it has was modified to the following `expr`

![Screenshot from 2023-05-12 10-59-21](https://github.com/canonical/grafana-agent-k8s-operator/assets/72372217/dee12076-20dc-4cdf-ab2c-3d4b0b413bba)

However, in Prometheus, `node_vmstat_oom_kill` metric does not have any timeseries with those labels.

![Screenshot from 2023-05-12 11-00-33](https://github.com/canonical/grafana-agent-k8s-operator/assets/72372217/e0df03c8-1096-4bee-9686-57d5f7a1e150)

So there's a label mismatch between actual metric from grafana-agent via node_exporter and from the alert rule.

### To Reproduce

1. Deploy / with existing COS-lite 
2. Deploy ubuntu charm and grafana-agent machine charm
3. relate ubuntu and grafana-agent; relate grafana-agent to COS-lite
4. Open Prometheus web UI, check the alert rules and inspect the `expr`

### Environment

Built from latest source @ 73543ef892912848c3babf062cbd17f8f7e20730, or `lastest/edge`

### Relevant log output

```shell
None
```

rgildein:
As we discussed today, this is not a bug, but a design decision which we want to improve.
But at the same time, we came to the conclusion that grafana-agent charm should offer the possibility of adding basic alert rules that are linked with principal application (JujuTopology is provided w/ information about principal application and not grafana-agent).

These rules should not be configurable and should have reasonable thresholds so that they fire only in situations when it is really critical. An example is an alert rule for memory usage, which has a threshold set to 95%. Such a threshold is high and therefore generic, so if any principal application provides another alert rule for memory, it will definitely be with lower threshold (more specific for application).

simskij:
@rgildein, you are mostly correct. 😅

There is a bug here where the alert rules get the `juju_charm` label, but the metrics and logs does not. We need to drop this label selector as it will mean that it will never match anything.

rgildein:
[...]

PR #244, opened 2023-09-07
overwrite the topology labels when recieving alerts over cos-agent

dstathis:
## Issue
The topology labels are wrong for alerts received over cos-agent.
Partially solves #190

## Solution
Overwrite the topology labels.

## Testing Instructions
1. Relate zookeeper, grafana-agent, and hardware-observer all to each other.
2. Relate grafana-agent to cos.
3. Check that alert rule expression all use zookeeper topology.

## Release Notes
Fixed and issue where the alert rules forwarded over cos-agent had the wrong labels.


sed-i (approved):
Is this something we could easily have a utest/stest for?

dstathis:
> Is this something we could easily have a utest/stest for?

Tests coming in another PR before I close the issue.

PR #246, opened 2023-09-08
insecure_skip_verify for localhost

### LINKED ISSUE #211: Allow COSAgentProvider to scrape public TLS metrics endpoint
### Enhancement Proposal

At the moment the machine `grafana-agent` subordinate charm assumes metrics endpoints would be served on localhost and does not expose scheme or tls_config for the endpoint. This can make it difficult to scrape apps that provide their own endpoints.

In the [Postgresql VM charm](https://github.com/canonical/postgresql-operator) we use [Patroni](https://github.com/zalando/patroni/) to manage the Postgresql cluster, which provides metrics as part of its REST API, however it is all served as a single endpoint. Exposing the API with TLS means also encrypting the metrics endpoint. The certificate used can also be self-signed.

To be able to consume that endpoint, we would need to be able to set the scheme for connection and either be able to use the host for which the certificate was signed or configure the agent to ignore the validation errors.

dstathis:
## Issue
Fixes #211

## Solution
Always use insecure_skip_verify for localhost

## Testing Instructions
1. Deploy grafana-agent
2. relate over cos-agent (to zookeeper for example)
3. check that insecure_skip_verify is in /etc/grafana-agent.yaml

## Release Notes
Grafana-agent can now scrape TLS endpoints provided over cos-agent.

### DIFF
--- a/lib/charms/grafana_agent/v0/cos_agent.py
+++ b/lib/charms/grafana_agent/v0/cos_agent.py
@@ -722,6 +722,8 @@ class COSAgentRequirer(Object):
                         "metrics_path": job["path"],
                         "static_configs": [{"targets": [f"localhost:{job['port']}"]}],
                         # We include insecure_skip_verify because we are always scraping localhost.
+                        # Even if we have the certs for the scrape targets, we'd rather specify the scrape
+                        # jobs with localhost rather than the SAN DNS the cert was issued for.
                         "tls_config": {"insecure_skip_verify": True},
                     }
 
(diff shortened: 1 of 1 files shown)


lib/charms/grafana_agent/v0/cos_agent.py
```diff
@@ -721,6 +721,8 @@ def metrics_jobs(self) -> List[Dict]:
                         "job_name": job["job_name"],
                         "metrics_path": job["path"],
                         "static_configs": [{"targets": [f"localhost:{job['port']}"]}],
+                        # We include insecure_skip_verify because we are always scraping localhost.
```
sed-i, line 724: Added more context in the comment. Makes sense?
```suggestion
                        # We include insecure_skip_verify because we are always scraping localhost.
                        # Even if we have the certs for the scrape targets, we'd rather specify the scrape
                        # jobs with localhost rather than the SAN DNS the cert was issued for.
```

sed-i (approved):

PR #247, opened 2023-09-11
increase snap client timeout to 20 seconds

neoaggelos:
## Issue

Test flakiness caused frequently in CI environments where resources are limited

## Solution

Increase snap client timeout to 20 seconds, seems to address the issue consistently for our case. Not sure if the observability team is happy to accept this patch or do something else instead. cc @sed-i

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -658,7 +658,7 @@ class SnapClient:
         socket_path: str = "/run/snapd.socket",
         opener: Optional[urllib.request.OpenerDirector] = None,
         base_url: str = "http://localhost/v2/",
-        timeout: float = 5.0,
+        timeout: float = 20.0,
     ):
         """Initialize a client instance.
 


lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -658,7 +658,7 @@ def __init__(
         socket_path: str = "/run/snapd.socket",
         opener: Optional[urllib.request.OpenerDirector] = None,
         base_url: str = "http://localhost/v2/",
-        timeout: float = 5.0,
+        timeout: float = 20.0,
```
sed-i, line 661: Changing it here (rather than https://github.com/canonical/operator-libs-linux/) would interfere with the `fetch-lib` mechanism.

I'd like to take a look at the flakiness issues encountered in https://github.com/canonical/charm-microk8s/pull/102.
cc @simskij 

BTW, would you accept this change at the source @benhoyt?

benhoyt: Yep, I'd accept a change (with justification) at the [source](https://github.com/canonical/operator-libs-linux/blob/20b8afd035e6f8a98390490d74676c150d746bcc/lib/charms/operator_libs_linux/v2/snap.py#L695)!

PR #249, opened 2023-09-14
Use operator_libs_linux.v2.snap

neoaggelos:
## Issue

Update to [latest version (v2) of operator_libs_linux](https://charmhub.io/operator-libs-linux/libraries/snap).

## Solution

```
charmcraft fetch-lib charms.operator_libs_linux.v2.snap
```

Also update current code that used v1.

This update is not included in #213

## Context

Replaces #247

Bumping a major version can be tricky, but the library methods used by the snap are not affected.


sed-i:
`Revision 17 of charm 'grafana-agent' released to edge`

PR #251, opened 2023-09-15
test to ensure alert rules are labeled correctly

dstathis:
## Issue
We don't have a test for this yet. Partially completes #190

## Solution
Add the test

## Testing Instructions
Not needed as this is itself a test.

## Release Notes
-

### DIFF
--- a/tests/scenario/test_machine_charm/test_alert_labels.py
+++ b/tests/scenario/test_machine_charm/test_alert_labels.py
@@ -1,0 +1,120 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import json
+
+import machine_charm
+from scenario import Context, PeerRelation, Relation, State, SubordinateRelation
+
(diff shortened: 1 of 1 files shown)


tests/scenario/test_machine_charm/test_alert_labels.py
```diff
@@ -0,0 +1,120 @@
...
+
+import json
+
+import machine_charm
+from scenario import Context, PeerRelation, Relation, State, SubordinateRelation
+
+from tests.scenario.helpers import get_charm_meta
+
+
+def test_metrics_alert_rule_labels(vroot):
```
sed-i, line 12: Could you add a docstring for the purpose of the test?

sed-i (approved):

lucabello (approved):

tests/scenario/test_machine_charm/test_alert_labels.py
dstathis: Done.

PR #257, opened 2023-09-26
Convert gagent shared code to a charm lib

sed-i:
## Issue
After splitting the hybrid repo into this charm and the machine charm, we need to figure out how to prevent unnecessary code duplication between the two.

## Solution
Convert the shared code in `grafana_agent.py` to a charm lib.

## Context
NTA.

## Testing Instructions
CI.

## Release Notes
Convert gagent shared code to a charm lib.


sed-i:
Rejected in favor of complete split. Closing.

PR #186, opened 2023-05-09
Add alert rules for disk utilization

rgildein:
- add alert rule for root disk space prediction reaching threshold
- add alert rule for root disk space reaching threshold
- add alert rule for checking readonly moutpoints
- add alert rule for XFS errors
- add alert rule for high read/write disk usage rate

## Context
Moving disk NRPE checks from [charm-nrpe](https://git.launchpad.net/charm-nrpe/tree/files/plugins).

## Testing Instructions
Tested with
```yaml
rule_files:
  - disk.rules

evaluation_interval: 1m

tests:
  # disk usage prediction
  - interval: 1m
    input_series:
      - series: 'used_disk_space{instance="test-model_1234_test-app_test-app/0", mountpoint="/"}'
        values: '10x60 20x60 30x60 40x60 50x60 60x60 70x60 80x60 90x60 100x60'
    promql_expr_test:
      - expr: used_disk_space
        eval_time: 1h
        exp_samples:
          - labels: '{__name__="used_disk_space", instance="test-model_1234_test-app_test-app/0", mountpoint="/"}'
            value: 10
      - expr: used_disk_space
        eval_time: 9h
        exp_samples:
          - labels: '{__name__="used_disk_space", instance="test-model_1234_test-app_test-app/0", mountpoint="/"}'
            value: 90
      - expr: predict_linear(used_disk_space{mountpoint=~"/"}[6h], 6*3600)
        eval_time: 4h
        exp_samples:
          - labels: '{instance="test-model_1234_test-app_test-app/0", mountpoint="/"}'
            value: 99.11525667844037
    alert_rule_test:
      - eval_time: 1h
        alertname: HostDiskSpaceFillsUp
        exp_alerts: []  # no alert
      - eval_time: 4h
        alertname: HostDiskSpaceFillsUp
        exp_alerts:
          - exp_labels:
              severity: warning
              mountpoint: /
              instance: test-model_1234_test-app_test-app/0
            exp_annotations:
              summary: "[Prediction] Host filesystem '/' is using 99% of the total space (instance test-model_1234_test-app_test-app/0)"
              description: >-
                Host filesystem '/' usage can potentially reach 99% of the total space.
                  VALUE = 99.11525667844037
                  LABELS = map[instance:test-model_1234_test-app_test-app/0 mountpoint:/]
                The 6-hour-ahead prediction is made as a linear regression from the last 60 minutes of data.
  # disk usage
  - interval: 1m
    input_series:
[...]

PR #258, opened 2023-09-27
Fix typo in disk.rules

chanchiwai-ray:
## Issue
The metric name `node_filesystem_size_bytes_bytes` appears to be incorrect due to typo. The correct metric name should be `node_filesystem_size_bytes`, and it can be verified from the [source code](https://github.com/prometheus/node_exporter/blob/master/collector/filesystem_common.go#L117) of node exporter.

This is a regression from #186 

## Solution
Fix the typo: `node_filesystem_size_bytes_bytes` -> `node_filesystem_size_bytes`

Screenshot from 2023-09-26 12-30-20
Screenshot from 2023-09-26 12-30-32

## Context
Prometheus alert rules

## Testing Instructions
N/A

## Release Notes
Fix typo in the Prometheus alert rules for disk

### DIFF
--- a/src/prometheus_alert_rules/disk.rules
+++ b/src/prometheus_alert_rules/disk.rules
@@ -4,7 +4,7 @@ groups:
   - record: used_disk_space
     expr: |
       100 * (1 - (
-        node_filesystem_free_bytes / node_filesystem_size_bytes_bytes)
+        node_filesystem_free_bytes / node_filesystem_size_bytes)
       )
   - alert: HostDiskSpaceFillsUp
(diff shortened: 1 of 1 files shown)


sed-i (approved):

sed-i:
Thanks @chanchiwai-ray!
Would you be able to do the same in https://github.com/canonical/grafana-agent-operator ?

chanchiwai-ray:
Sure, I was not aware there's another repo for grafana-agent

rgildein (approved):
Thanks Ray.

PR #274, opened 2024-01-31
feat: Adding the Terraform module

gatici:
## Feature

This PR adds the Terraform module for the charm. Hence, the charm could be deployed using Terraform.


terraform/CONTRIBUTING.md
```diff
@@ -0,0 +1,86 @@
...
+### Prerequisites
+
+Make sure the following software and tools are installed in the development
+environment.
+
+- `microk8s`
+- `juju`
+- `terraform`
+
+### Prepare Development Environment
```
sed-i, line 14: Should we link to a docs page that covers this topic? Options:
- https://discourse.charmhub.io/t/set-up-your-development-environment/7379
- https://juju.is/docs/sdk/dev-setup

sed-i (commented):
Thanks @gatici!
I just skimmed briefly for now.
My impression was that cos in its entirety would be one terraform module.
Is there a particular use case this PR is addressing?

gatici:
> Thanks @gatici! I just skimmed briefly for now. My impression was that cos in its entirety would be one Terraform module. Is there a particular use case this PR is addressing?

Hi Leon,
Yes, We need the cos bundle as a Terraform module, that would be perfect if we have it as soon as possible. Besides, Grafana-agent-k8s is part of sdcore bundle. Please see: https://github.com/canonical/sdcore-bundles/blob/552ac796aea231f7e81d2446019e527a7fcd0f12/render_bundle/bundles.py#L42 So we need to deploy Grana charm by Terraform. A Terraform module may include another TF module and could be a part of a higher level TF module. So, creating Grafana TF module could be a starting point.

terraform/CONTRIBUTING.md
gatici: I checked the links that you provided. It seems they do not exactly match with our  requirements. Thank you for the idea. We can add a discourse page about development environment for Terraform.

gatici: I created a discourse page:  https://discourse.charmhub.io/t/set-up-your-development-environment-with-microk8s-for-juju-terraform-provider/13109

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,30 @@
...
+    paths:
+      - '**.tf'
+
+jobs:
+  build:
+    name: Terraform Checks and Plans
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository to the runner
+        uses: actions/checkout@v3
```
Gmerold, line 14:
```suggestion
        uses: actions/checkout@v4
```

terraform/main.tf
```diff
@@ -0,0 +1,42 @@
+resource "juju_application" "grafana-agent-k8s" {
+  name  = "grafana-agent-k8s"
+  model = var.model_name
+
+  charm {
+    name    = "grafana-agent-k8s"
+    channel = var.channel
+    base    = "ubuntu@22.04"
```
Gmerold, line 8: In Grafana's `charmcraft.yaml` we have:
```
bases:
  - build-on:
      - name: "ubuntu"
        channel: "20.04"
    run-on:
      - name: "ubuntu"
        channel: "20.04"
```
Do we need to specify the base at all here?

terraform/main.tf
```diff
@@ -0,0 +1,42 @@
...
+  }
+  config = var.grafana-config
```
Gmerold, line 10: I feel like decoupling `var.grafana-config ` into individual config params would be nicer. Having separate var for each config option would allow us leveraging the `description` field for each var. WDYT?

terraform/main.tf
```diff
@@ -0,0 +1,42 @@
...
+
+  application {
+    name     = juju_application.grafana-agent-k8s.name
+    endpoint = "logging-consumer"
+  }
+
+  application {
+    offer_url = var.logging_offer_url
+  }
+}
```
Gmerold, line 41: As already proposed in [this PR](https://github.com/canonical/sdcore-nms-k8s-operator/pull/46#discussion_r1481148685), I think we shouldn't be defining integrations in the base modules. 
My proposal is to model the `required` and `provided` integrations (relations) in the `outputs.tf` instead to be able to reference them in a higher level module.

terraform/terraform.tfvars
```diff
@@ -0,0 +1,10 @@
+# Mandatory Config Options
```
Gmerold, line 1: I think this base module is not intended to be used alone, so we probably don't need this file.

terraform/variables.tf
```diff
@@ -0,0 +1,28 @@
...
+  default     = ""
+}
+
+variable "channel" {
+  description = "The channel to use when deploying a charm "
+  type        = string
+  default     = "latest/stable"
+}
+
+variable "grafana-config" {
```
Gmerold, line 13: Please see the proposal regarding config options above.

terraform/main.tf
gatici: When I check the charmcraft.yaml, it uses ubuntu 22.04:
```
type: charm
bases:
  - build-on:
    - name: "ubuntu"
      channel: "22.04"
      architectures: ["amd64"]
    run-on:
    - name: "ubuntu"
      channel: "22.04"
      architectures: ["amd64"]
```

gatici: I think w can get rid of base.

terraform/main.tf
gatici: Writing each individual config param one by one in the variables.tf seems not necessary. Writing `description` in the variables is just  repeating  the config.yaml. Using a general config map also makes the modules' maintenance easier.

terraform/main.tf
gatici: Agreed.

terraform/variables.tf
gatici: I commented above. Besides, as I remember from one of the dailies, expanding config-options was not our choice.

.github/workflows/terraform.yaml
gatici: Done

terraform/main.tf
```diff
@@ -0,0 +1,42 @@
+resource "juju_application" "grafana-agent-k8s" {
+  name  = "grafana-agent-k8s"
```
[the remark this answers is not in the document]
gatici: Done

terraform/terraform.tfvars
gatici: Removed

terraform/main.tf
Gmerold: True. I agree with all of your arguments, but at the same time it requires potential user of this module to be familiar with the structure of charmed operators - to fill in the config for the Terraform module, they would have to know they should check the `config.yaml` first.
Similarly, I think that the charm relations should be modeled in the `output.tf`. Each of them separately, to give the user a clear idea of what public interfaces this module supports. I wouldn't rely on the assumption that the user knows where to find things. 
[Here](https://github.com/canonical/cos-configuration-k8s-operator/pull/81) is an example of how I envision the base Terraform module. Let me know what you think.

terraform/main.tf
gatici: Done

terraform/main.tf
gatici: It is stated several times here that the modules are not planned to be used standalone and we are not expecting the consumers for a single module. I think can give reference in README.md about how to find config details.
There is a benefit of using the references as it's easy to maintain. We do not need to update those information twice if there is change.
I like the README.md which explains TF module structure and including a sample how to integrate sections.

Gmerold: I don't think breaking the config into individual items and using the module in separation are anyhow related. 
Anyway, I see benefits of both approaches and don't really have a very strong opinion on this one. I'd be interested in hearing from others.

gatici: Agree. We need the team's opinions here.

Gmerold: And by the way, one more argument for having configs separated is that if we want to change just one config option and leave the rest of them with the default values, we can. With monolithic config - we can't ;)
Looks like my opinion on this one gets stronger after all :rofl:

gatici: According to the tests results, we get the same behaviour. Only the (key: value) pairs which are set in the monolithic config has changed, the other config options are not affected (using their default values).

.github/workflows/terraform.yaml
Gmerold: That's the `checkout` action, not `setup-terraform`.

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Getting Started
+
+### Prerequisites
+
+The following software and tools needs to be installed and should be running in the local environment. Please [set up your environment](https://discourse.charmhub.io/t/set-up-your-development-environment-with-microk8s-for-juju-terraform-provider/13109) before deployment.
+
+- `microk8s`
+- `juju 3.x`
+- `terrafom`
```
Gmerold, line 13:
```suggestion
- `Terraform`
```

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+This Grafana-agent-k8s Terraform module aims to deploy the [grafana-agent-k8s charm](https://charmhub.io/grafana-agent-k8s) via Terraform.
+
+## Getting Started
+
+### Prerequisites
+
+The following software and tools needs to be installed and should be running in the local environment. Please [set up your environment](https://discourse.charmhub.io/t/set-up-your-development-environment-with-microk8s-for-juju-terraform-provider/13109) before deployment.
+
+- `microk8s`
+- `juju 3.x`
```
Gmerold, line 12: Is Juju 3.x required? Terraform provider works with Juju `2.9.33+` and I don't think this charm requires Juju 3 specifically, does it?

terraform/README.md
```diff
@@ -0,0 +1,65 @@
+# Grafana-agent-k8s Terraform Module
+
+This Grafana-agent-k8s Terraform module aims to deploy the [grafana-agent-k8s charm](https://charmhub.io/grafana-agent-k8s) via Terraform.
+
+## Getting Started
+
+### Prerequisites
```
Gmerold, line 7: I think it would be good to mention, that the Juju controller needs to be bootstrapped onto the cluster before we can use this module.

terraform/README.md
```diff
@@ -0,0 +1,65 @@
+# Grafana-agent-k8s Terraform Module
+
+This Grafana-agent-k8s Terraform module aims to deploy the [grafana-agent-k8s charm](https://charmhub.io/grafana-agent-k8s) via Terraform.
```
Gmerold, line 3: How about we mention that the module uses Juju Terraform provider, as we did for the SD-Core modules?

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+```text
+module "grafana-agent-k8s" {
+  source = "git::https://github.com/canonical/grafana-agent-k8s-operator//terraform"
+  
+  model_name = "juju_model_name"
+  # Optional Configurations
+  # channel                        = "put the Charm channel here" 
+  # grafana-config = {
+  #   tls_insecure_skip_verify = "put True not to skip the TLS verification"
+  # }
```
Gmerold, line 35: Let's roll back to the monolithic config, as we agreed.

terraform/main.tf
```diff
@@ -0,0 +1,16 @@
...
+
+resource "juju_application" "grafana-agent-k8s" {
+  name  = var.app_name
+  model = var.model_name
+
+  charm {
+    name    = "grafana-agent-k8s"
+    channel = var.channel
+  }
+  config = var.grafana_config
```
Gmerold, line 12: I think `grafana_agent_config` or just `config` would be more suitable here.

terraform/README.md
Gmerold, line 35:
```suggestion
  config = var.config
```

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+}
+```
+
+Please see the link to customize the Grafana configuration variables if needed.
+
+- [Grafana configuration option](https://charmhub.io/grafana-agent-k8s/configure)
```
Gmerold, line 41: Since we'll be using monolithic config, I think we don't need this - it's already mentioned in the `variables.tf`. WDYT?

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+Create the integrations, for instance:
+
+```text
+resource "juju_integration" "amf-metrics" {
```
Gmerold, line 46:
```suggestion
resource "juju_integration" "grafana-agent-metrics" {
```

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+  model = var.model_name
+
+  application {
+    name     = module.amf.app_name
+    endpoint = module.grafana.metrics_endpoint
```
Gmerold, line 51:
```suggestion
    name     = module.grafana-agent-k8s.app_name
    endpoint = module.grafana-agent-k8s.metrics_endpoint
```

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  application {
+    name     = module.grafana.app_name
+    endpoint = module.grafana.metrics_endpoint
```
Gmerold, line 56:
```suggestion
    name     = module.prometheus.app_name
    endpoint = module.prometheus.metrics_endpoint
```

Gmerold: Typically it's the Prometheus that provides the `metrics-endpoint` integration.

terraform/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+  default     = "grafana"
+}
+
+variable "channel" {
+  description = "The channel to use when deploying a charm "
+  type        = string
+  default     = "latest/stable"
+}
+
+variable "grafana_config" {
```
Gmerold, line 22: Please see the comment above - I'd suggest `grafana_agent_config` or simply `config`.

terraform/variables.tf
```diff
@@ -0,0 +1,25 @@
...
+  description = "Additional configurations for the Grafana. Please see the available options: https://charmhub.io/grafana-agent-k8s/configure"
```
Gmerold, line 23:
```suggestion
  description = "Additional configurations for the Grafana Agent. Please see the available options: https://charmhub.io/grafana-agent-k8s/configure"
```

terraform/outputs.tf
```diff
@@ -0,0 +1,52 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+output "app_name" {
+  description = "Name of the deployed application."
+  value       = juju_application.grafana-agent-k8s.name
+}
+
```
Gmerold, line 8: Just a suggestion, but I've been adding comments to separate required integration from provided integrations. I think this improves readability.

terraform/README.md
```diff
@@ -0,0 +1,65 @@
+# Grafana-agent-k8s Terraform Module
+
+This Grafana-agent-k8s Terraform module aims to deploy the [grafana-agent-k8s charm](https://charmhub.io/grafana-agent-k8s) via Terraform.
+
+## Getting Started
```
Gmerold, line 5: In general, I'm not sure if this section is needed, since the module is intended to be imported into a higher level module. WDYT?

Gmerold (changes requested):

terraform/main.tf
Gmerold: Discussed and agreed to keep the config as a monolith. 
Let's close this convo (I can't do it ;) ).

terraform/variables.tf
Gmerold: Yes, let's close this.

terraform/README.md
simskij: All of COS requires 3.x, except grafana-agent and cos-proxy.

.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,30 @@
+name: Check the Terraform Module
```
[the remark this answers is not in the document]
gatici: I will add that.

terraform/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Prerequisites
+
+The following software and tools needs to be installed and should be running in the local environment. Please [set up your environment](https://discourse.charmhub.io/t/set-up-your-development-environment-with-microk8s-for-juju-terraform-provider/13109) before deployment.
+
+- `microk8s`
```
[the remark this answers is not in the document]
gatici: This is updated as `Kubernetes Cluster`.

terraform/README.md
gatici: I will remove the version info as it could may vary depending on application type.

terraform/README.md
gatici: https://discourse.charmhub.io/t/set-up-your-development-environment-with-microk8s-for-juju-terraform-provider/13109 link includes steps to prepare the environment (including Juju controller bootstrap) so we did not write it again.

.github/workflows/terraform.yaml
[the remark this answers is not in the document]
gatici: Done

terraform/README.md
gatici: Done

terraform/README.md
gatici: Done

terraform/README.md
gatici: Added.

terraform/README.md
gatici: Done

terraform/variables.tf
gatici: Done

terraform/README.md
gatici: This is removed.

gatici:
We will store the traefik-k8s TF module under this repository temporarily https://github.com/canonical/terraform-juju-sdcore-k8s/pull/4/files. When you had a decision, we can move to it's permanent place.

PR #282, opened 2024-03-21
Fix SSL Cert Issue in Certificate Transfer

IbraAoad:
## Issue
Complementary to the fix of https://github.com/canonical/loki-k8s-operator/issues/360

## Solution
This PR points the `_ca_folder_path` to the correct path to solve the `certificate signed by unknown authority issue` when `gagent` is related over the `certificate_transfer` interface

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -249,7 +249,7 @@ class GrafanaAgentK8sCharm(GrafanaAgentCharm):
         Args:
             cmd: Command to be run.
         """
-        self._container.exec(cmd)
+        self._container.exec(cmd).wait()
 
     @property
     def tracing_endpoint(self) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


mmkay (approved):

src/charm.py
```diff
@@ -249,7 +249,7 @@ def run(self, cmd: List[str]):
         Args:
             cmd: Command to be run.
         """
-        self._container.exec(cmd)
+        self._container.exec(cmd).wait()
```
mmkay, line 252: Do we need the `.wait()`? The other PR doesn't have it.

IbraAoad: Yeah doesn't really work without it, The other PR exec via subprocess so we don't need it there

PR #286, opened 2024-04-09
Use helper for mandatory relation pairs

sed-i:
## Issue
Mandatory relation pairs were modeled incorrectly.

## Solution
Use helper for mandatory relation pairs.

In tandem with:
- https://github.com/canonical/cos-lib/pull/25
- https://github.com/canonical/grafana-agent-operator/pull/85

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,8 +4,7 @@
 # FIXME: Packing the charm with 2.2.0+139.gd011d92 will not include dependencies in PYDEPS key:
 # https://chat.charmhub.io/charmhub/pl/wngp665ycjnb78ar9ojrfhxjkr
 # That's why we are including cosl here until the bug in charmcraft is solved
-#cosl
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -4,7 +4,8 @@
 # FIXME: Packing the charm with 2.2.0+139.gd011d92 will not include dependencies in PYDEPS key:
 # https://chat.charmhub.io/charmhub/pl/wngp665ycjnb78ar9ojrfhxjkr
 # That's why we are including cosl here until the bug in charmcraft is solved
-cosl
+#cosl
+git+https://github.com/canonical/cos-lib@feature/mandatory-relation-pairs
```
Abuelodelanada, line 8: Why do not we wait for cosl to be released to pypi so we do not have to create another PR ?

Abuelodelanada (approved):

requirements.txt
sed-i: Yep, done!

PR #297, opened 2024-05-03
Add ARM build

lucabello:


sed-i (approved):

sed-i:
@taurus-forever, when this merges there should be an arm release on charmhub.

PR #310, opened 2024-08-06
Tracing support

PietroPasotti:
This PR adds tracing support for grafana agent.

The goal is to configure grafana-agent in such a way that, when CMR'd with a Tempo backend (for example in cos-lite), grafana will forward any traces it receives to Tempo.


src/UNKNOWN.egg-info/PKG-INFO
mmkay, line 1: This too seems like a spillover from things that should be in `.gitignore` but aren't?

src/grafana_agent.py
```diff
@@ -78,12 +87,30 @@ class GrafanaAgentCharm(CharmBase):
...
     _grpc_listen_port = 3600
-    # TODO Change to a more suitable location once the snap gets access (#216).
+
     _cert_path = "/tmp/agent/grafana-agent.pem"
     _key_path = "/tmp/agent/grafana-agent.key"
     _ca_path = "/usr/local/share/ca-certificates/grafana-agent-operator.crt"
     _ca_folder_path = "/usr/local/share/ca-certificates"
 
+    # mapping from tempo-supported receivers to the receiver ports to be opened on the grafana-agent host
+    # to ingest traces for them. Note that we 'support' more receivers here than tempo currently does, so
```
mmkay, line 97: Nit: Technically not anymore after canonical/tempo-k8s-operator#149.

src/grafana_agent.py
```diff
@@ -198,23 +258,81 @@ def __init__(self, *args):
...
+        self._update_status()
+        self._update_tracing_provider()
 
-    def _on_mandatory_relation_event(self, _event=None):
-        """Event handler for any mandatory relation event."""
+    def _on_tracing_endpoint_removed(self, _event) -> None:
+        """Event handler for the tracing endpoint-removed event."""
         self._update_config()
         self._update_status()
+        self._update_tracing_provider()
```
mmkay, line 315: Idea: we seem to be having this sequence of `_update_config()`, `_update_status()`, `_update_tracing_provider()` in several following methods. Could we extract it to a single update method that would be calling all that?

grafana_dashboards/node-exporter-full.json
mmkay, line 1: Probably a `git add` leftover.

loki_alert_rules/grafana_agent_high_rate.rule
mmkay, line 1: These likely can go too.

prometheus_alert_rules/.wokeignore
mmkay, line 1: ...and this whole folder.

mmkay (approved):
A bit of housekeeping complaints, other than that LGTM!

src/grafana_agent.py
PietroPasotti: +1 to that, but refactoring this code is out of scope for this PR. 
I'd like to move the status management in `collect-unit-status` where it should be, and then indeed, update_config should take care of updating all configs. I didn't want to break with the existing pattern for this specific PR

grafana_dashboards/node-exporter-full.json
PietroPasotti: uh yes, my local must have had a very old main branch

PR #333, opened 2024-12-02
Rename the title of the node-exporter dashboard

cbartz:
## Issue

There is a dashboard title collision with the one in the grafana-agent machine charm.

## Solution

Rename the title as suggested in https://github.com/canonical/observability/pull/206/files#r1863947198
uid collision will be addressed later as indicated in https://github.com/canonical/observability/pull/206.

### DIFF
--- a/src/grafana_dashboards/node-exporter-full.json
+++ b/src/grafana_dashboards/node-exporter-full.json
@@ -23194,7 +23194,7 @@
   },
   "timezone": "browser",
   "title": "System Resources (container)",
-  "uid": "rYdddlPWk",
+  "uid": "rYdddlPWk-1",
   "version": 10,
   "weekStart": ""
 }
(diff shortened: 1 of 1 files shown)


sed-i (approved):

src/grafana_dashboards/node-exporter-full.json
```diff
@@ -23193,8 +23193,8 @@
     ]
   },
   "timezone": "browser",
-  "title": "System Resources",
+  "title": "System Resources (container)",
   "uid": "rYdddlPWk",
```
sed-i, line 23197: UID is overlapping with the other charm. How about:
```suggestion
  "uid": "rYdddlPWk-1",
```

sed-i (approved):
Tested manually, works:
```bash
curl http://admin:O3LQxaje62Kt@10.1.157.124:3000/api/search   
```
```json
[
  {
    "id": 1,
    "uid": "rYdddlPWk-1",
    "title": "System Resources (container)",
    "uri": "db/system-resources-container",
    "url": "/d/rYdddlPWk-1/system-resources-container",
    "slug": "",
    "type": "dash-db",
    "tags": [
      "linux"
    ],
    "isStarred": false,
    "sortMeta": 0
  }
]
```

PR #340, opened 2025-01-15
fix: apply HostDiskSpace alert to all relevant mountpoints

### LINKED ISSUE #331: HostDiskSpace alert doesn't trigger with mountpoints below `/`
### Bug Description

If we have a mountpoint for anything under `/` the alert doesn't trigger when the disk usage of this mountpoint is above the threshold. 

That seems to be due the HostDiskSpace metric uses a regex matching plain `/` and not everything within it:
`used_disk_space{mountpoint=~"/"} > 90`

A lazy approach would be changing to `mountpoint=~"/.*"`. That way, everything below `/` would match the alert.

### To Reproduce

On any monitored system
```bash
# create a dummy file
fallocate -l10G /var/tmp/host_disk_space
# get a free loop device
losetup -f 
# use the free loop device
losetup /dev/loop9 /var/tmp/host_disk_space
# create the file system
mkfs -t ext4 /dev/loop9
# mount
mount /dev/loop9 /mnt/
# fill up the mountpoint
fallocate -l9G /mnt/fillup
```
No alerts are generated that way.

### Environment

n/a

### Relevant log output

```shell
n/a
```

### Additional context

_No response_

lucabello:
## Issue
Closes #331, together with [grafana-agent#233](https://github.com/canonical/grafana-agent-operator/pull/233).

## Solution
Update the alert to include all mountpoints (removing the `mountpoint~="/"` filter), but excluding irrelevant mountpoints:
- `fstype: fuse.lxcfs` is always NaN (confirmed by the LXD team)
- `fstype: fuse.snapfuse` is always 100%
- `/run`, `/proc`, and `/dev` are not relevant

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@ tests/integration/*-tester/lib/
 .env
 cos-tool*
 *.egg-info
-.charm_tracing_buffer.raw
 # Terraform
 *.tfstate
 *.tfstate.*
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ tests/integration/*-tester/lib/
 .env
 cos-tool*
 *.egg-info
+.charm_tracing_buffer.raw
 # Terraform
 *.tfstate
 *.tfstate.*
--- a/src/prometheus_alert_rules/disk.rules
+++ b/src/prometheus_alert_rules/disk.rules
@@ -20,7 +20,7 @@ groups:
         The 6-hour-ahead prediction is made as a linear regression from the last 60 minutes of data.
   - alert: HostDiskSpace
     # some fstype are excluded because they are not relevant; see https://github.com/canonical/grafana-agent-k8s-operator/pull/340 for details
-    expr: used_disk_space{fstype!~"fuse\.lxcfs|fuse\.snapfuse", mountpoint!~"/run.*|/proc|/dev"} > 90
+    expr: used_disk_space{fstype!~"fuse\\.lxcfs|fuse\\.snapfuse", mountpoint!~"/run.*|/proc|/dev"} > 90
     for: 0m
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -10,6 +10,7 @@ tests/integration/*-tester/lib/
 .env
 cos-tool*
 *.egg-info
+.charm_tracing_buffer.raw
```
sed-i, line 13: What's that?
Does it have to reside in the charm dir?
I.e. if this is a reasonable issue to raise on the charm tracing lib then we probably shouldn't have it here?

src/prometheus_alert_rules/disk.rules
```diff
@@ -19,7 +19,7 @@ groups:
           LABELS = {{ $labels }}
         The 6-hour-ahead prediction is made as a linear regression from the last 60 minutes of data.
   - alert: HostDiskSpace
-    expr: used_disk_space{mountpoint=~"/"} > 90
+    expr: used_disk_space{fstype!~"fuse.lxcfs|fuse.snapfuse", mountpoint!~"/run.*|/proc|/dev"} > 90
```
sed-i, line 22: - We should probably escape the `.` in `fuse.lxcfs` and `fuse.snapfuse`?
- Add comment explaining why they're excluded?

sed-i (approved):

src/prometheus_alert_rules/disk.rules
lucabello: Done

.gitignore
lucabello: After checking with the tracing team, I added a fixture to move that file somewhere else.

sed-i (approved):

PR #355, opened 2025-03-10
Added `storage` directives to terraform configuration

### LINKED ISSUE #351: Expose `storage directives` as input for terraform module
### Enhancement Proposal

This charm's TF module is currently not exposing the [`storage directives`](https://juju.is/docs/juju/storage-constraint) as an input variable. This is not pressing, but a nice to have.

swetha1654:
## Issue

Resolves #351 

## Solution

Added [storage_directives](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#storage_directives-1) as an input variable, which defaults to 1 GB.

## Context

This is an _optional_ Juju application TF resource config, meaning that this feature is a "nice to have".

## Testing Instructions

```
terraform plan
terraform apply
```
and check the storage with `juju storage`:
```
Unit       Storage ID  Type        Pool        Size     Status    Message
grafana/0  data/0      filesystem  kubernetes  1.0 GiB  attached  Successfully provisioned volume pvc-8690a030-fd92-4cfc-b5a5-51156d033585
```

## Upgrade Notes

This is backwards compatible


terraform/variables.tf
```diff
@@ -39,4 +39,10 @@ variable "units" {
   description = "Number of units"
   type        = number
   default     = 1
-}
\ No newline at end of file
+}
+
+variable "grafana_data-storage-size" {
```
MichaelThamm, line 44: Is there a reason you decided to not use underscores in the var name?
```suggestion
variable "grafana_data_storage_size" {
```

terraform/README.md
```diff
@@ -21,6 +21,7 @@ The module offers the following configurable inputs:
 | `model_name`| string | Name of the model that the charm is deployed on |  |
 | `revision`| number | Revision number of the charm name | null |
 | `units`| number | Number of units to deploy | 1 |
+| `grafana_data-storage-size`| number | Number of units to deploy | 1 |
```
MichaelThamm, line 24: Please fix the description here, e.g. include `<count>,<size>,<label>=<pool>` and a mention that this sets the `Storage constraint (directive)`. Also, a comment about what unit (if we still think hard-coding the unit is a good idea) is being configured.

terraform/main.tf
```diff
@@ -9,4 +9,7 @@ resource "juju_application" "grafana_agent" {
   }
   units  = var.units
   config = var.config
+  storage_directives = {
+    data = "${var.grafana_data-storage-size}GB,kubernetes"
```
MichaelThamm, line 13: I wonder if we should instead accept the `<count>,<size>` as input rather than hardcoding `GB`. Otherwise, I fear that we limit the user too much. Alternatively, we can keep it but update the description to limit this to GB

MichaelThamm (changes requested):

MichaelThamm:
After discussions about Terraform spec regarding storage directives, it is not clear what the direction is yet. We should resolve the design first before merging this PR.

terraform/variables.tf
swetha1654: nope, just a typo.

terraform/README.md
swetha1654: changed it to <count><size>

terraform/main.tf
swetha1654: changed it to {count}{size}

terraform/variables.tf
```diff
@@ -39,4 +39,10 @@ variable "units" {
   description = "Number of units"
   type        = number
   default     = 1
-}
\ No newline at end of file
+}
+
+variable "grafana_data_storage_size" {
```
MichaelThamm, line 44: According to the spec:

<b style="font-weight:normal;" id="docs-internal-guid-b0ae7e34-7fff-8fd9-fca8-a6665610170d"><div dir="ltr" style="margin-left:0pt;" align="center">
Name | Type | Default Value | Description
-- | -- | -- | --
storage | map(string) | {} | Map of storage used by the application

</div></b>

So we should call this variable "storage" whith the following type and default.

terraform/variables.tf
```diff
@@ -39,4 +39,10 @@ variable "units" {
...
+  description = "Grafana agent charm: set the size of the data volume. The default is 1 GB."
+  type        = string
+  default     = 1GB
+}
```
MichaelThamm, line 48: According to the spec:

<b style="font-weight:normal;" id="docs-internal-guid-b0ae7e34-7fff-8fd9-fca8-a6665610170d"><div dir="ltr" style="margin-left:0pt;" align="center">
Name | Type | Default Value | Description
-- | -- | -- | --
storage | map(string) | {} | Map of storage used by the application

</div></b>


```suggestion
variable "storage" {
  description = "Map of storage used by the application."
  type        = map(string)
  default     = {}
}
```

MichaelThamm: What is interesting to investigate here is what the TF provider tell Juju to provision when the default is used: `{}`. If we are opintionated on this by defining a static string like:

```
  default     = {
    data = "kubernetes,1,1024M"
  }
```

terraform/README.md
```diff
@@ -21,6 +21,7 @@ The module offers the following configurable inputs:
 | `model_name`| string | Name of the model that the charm is deployed on |  |
 | `revision`| number | Revision number of the charm name | null |
 | `units`| number | Number of units to deploy | 1 |
+| `grafana_data_storage_size`| number | Size of the data volume | 1GB |
```
MichaelThamm, line 24: FIXME: Update the default with the value we decide in this PR
```suggestion
| `storage`| map(string) | Map of storage used by the application. | {} |
```

terraform/main.tf
```diff
@@ -9,4 +9,7 @@ resource "juju_application" "grafana_agent" {
   }
   units  = var.units
   config = var.config
+  storage_directives = {
+    data = "${var.grafana_data-storage-size},kubernetes"
+  }
```
MichaelThamm, line 14:
```suggestion
  storage_directives = var.storage
```

MichaelThamm (changes requested):

terraform/variables.tf
swetha1654: I guess we don't need to provide a default value in the terraform files, juju takes care of it. If required i can add a line in the description mentioning that the default value is 1 GB if not provided. Let me know!

swetha1654:
@MichaelThamm i see that the Check libraries CI failing with the error `Failed to unlock keyring`. Might that have something to do with the an expired CharmHub token?

swetha1654:
@MichaelThamm i think the PR is good to go :)

PR #375, opened 2025-06-02
feat: remove HighLogVolume alert rule

sed-i:
## Issue
The logql alert expression `count_over_time(({%%juju_topology%%})[30s]) > 100` addresses logs emitted by grafana-agent itself. However, `100` is an arbitrary threshold. For example, grafana-agent will emit a log line for every rotated file, which can total [high per-minute values](https://github.com/canonical/grafana-agent-operator/issues/153). 

## Solution
Remove the alert. In tandem with:
- https://github.com/canonical/grafana-agent-operator/pull/272.

Fixes https://github.com/canonical/grafana-agent-operator/issues/150

## Context
- Changing this alert to account a "moving window context", e.g. `rate([5m]) > 10*rate([5m] offset 5m)` is slightly better, but the factor, e.g. `10` is still arbitrary and does not capture spiking.

### DIFF
--- a/src/loki_alert_rules/high_error_rate.rule
+++ b/src/loki_alert_rules/high_error_rate.rule
@@ -1,8 +1,0 @@
-alert: HostHighLogErrorRate
-expr: count_over_time({job="varlogs"} |= "error" [1h]) > 100
-for: 0m
-labels:
-  severity: warning
-annotations:
-  summary: High error rate in logs (instance {{ $labels.instance }})
-  description: "High error rate in logs\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"


MichaelThamm (dismissed):
According to [this comment](https://github.com/canonical/grafana-agent-operator/pull/272#issuecomment-2933932430) in the tandem PR, we should not remove the `high_error_rate.rule` because it is not creating noise in production. Also, [this tandem PR fixes a unit test](https://github.com/canonical/grafana-agent-operator/pull/272/), which is missing in this PR. I think this PR will become obsolete once the other one is merged.

sed-i:
> According to [this comment](https://github.com/canonical/grafana-agent-operator/pull/272#issuecomment-2933932430) in the tandem PR, we should not remove the `high_error_rate.rule` because it is not creating noise in production. Also, [this tandem PR fixes a unit test](https://github.com/canonical/grafana-agent-operator/pull/272/), which is missing in this PR. I think this PR will become obsolete once the other one is merged.

Since the `HostHighLogErrorRate` alert rule was mistakenly left out of the VM charm, and does not apply to the k8s charm, and has a typo -

```diff
-varlogs
+varlog
```

-- I am going to merge this as is, removing both alert rules from this k8s charm.

Issue #382, opened 2025-06-17
When multiple identical remote_write endpoints are provided the workload crashes


lucabello:
### Bug Description

When using a scaled-up Mimir coordinator, the grafana-agent adds all the Mimir URLs in its `remote_write` configuration.

However, if Mimir is behind an ingress, the urls will all be the same, and this replication makes grafana-agent crash.

An easy solution is to deduplicate the section in grafana-agent.

### To Reproduce

Deploy a Mimir cluster with scaled-up coordinator and ingress, then relate grafana-agent to remote-write to it.

### Environment

-

### Relevant log output

```shell
-
```

PR #384, opened 2025-06-18
fix: deduplicate remote_write addresses to avoid crashes

lucabello:
## Issue
Addresses #382 for track **2** (main).

## Solution
Addresses in the `remote_write` sections are now deduplicated to avoid crashes.

## Context
This has been discovered and addressed in the context of PC testing. Note that this only affects remote_write sections of the config, so Loki and Tempo are unaffected.

## Testing Instructions
Review the PR and run the unit test, or: manually try to deploy the charm with a scaled-up Mimir coordinator that's using Traefik.

### DIFF
--- a/tests/unit/test_endpoint_deduplication.py
+++ b/tests/unit/test_endpoint_deduplication.py
@@ -5,7 +5,7 @@ import json
 
 import yaml
 from helpers import k8s_resource_multipatch, patch_lightkube_client
-from ops.testing import Container, Context, Relation, State, Exec
+from ops.testing import Container, Context, Exec, Relation, State
 
 from charm import GrafanaAgentK8sCharm
 from grafana_agent import CONFIG_PATH


tests/unit/test_endpoint_deduplication.py
```diff
@@ -0,0 +1,57 @@
...
+        ],
+        config={
+            "forward_alert_rules": True,
+        },
+    )
+
+    # WHEN the charm processes endpoints with duplicates
+    ctx = Context(charm_type=GrafanaAgentK8sCharm)
+
+    with ctx(ctx.on.config_changed(), state) as mgr:
```
PietroPasotti, line 38: you don't need to use the context manager for this test. 
```suggestion
    output_state = ctx.run(ctx.on.config_changed(), state)
    <unindent the rest>
```

PietroPasotti (approved):

tests/unit/test_endpoint_deduplication.py
lucabello: Great, done!

lucabello:
We should likely do this change in the remote write library instead.

PR #411, opened 2025-10-01
fix: restart pebble on cert write

### LINKED ISSUE #403: Agent doesn't receive certificate over `receive-ca-cert` in CMR
### Bug Description

Our docs on [setting up TLS encryption](https://documentation.ubuntu.com/observability/how-to/configure-tls-encryption/) show a `ca-model` containing a certificates provider, relating to a `grafana-agent` in a different model over the `certificate_transfer` interface.

However, this setup doesn't work: the certificate successfully travels over relation data, but for some reason the charm doesn't save it to disk - and thus doesn't trust it.

This bug happens for both grafana-agent k8s and machine.

### To Reproduce

```bash
# on a k8s controller
## Setup CA model
juju add-model ca-model
juju deploy self-signed-certificates ssc
juju offer ssc:send-ca-cert
## Setup grafana-agent model
juju add-model observable-model
juju deploy grafana-agent-k8s agent
juju consume <controller>:admin/ca-model.ssc
juju relate agent:receive-ca-cert ssc
```

### Environment

-

### Relevant log output

```shell
-
```

### Additional context

_No response_

sinapah:
## Issue

This PR fixes #403. 
As the last comment in the issue explains, the issue happens because the agent service is not restarted after the certificate is written to disk. 

## Solution

Restarts Pebble's agent service when the new certs are written to disk.
This PR also does `charmcraft fetch-lib` and bumps `cosl`.


lucabello (approved):
Weird that the tls_certificates relation doesn't require a restart but certificate_transfer does, I wonder if we're missing something

